PR #23, opened 2019-07-23
add slice and zeropad operators

jinzhongxiao:


jaybdub:
Hi jinzhongxiao,

Thanks for this contribution!  I think the first step in merging this would be to add a few module tests to ensure the functionality.  [mean.py](https://github.com/NVIDIA-AI-IOT/torch2trt/blob/master/torch2trt/converters/mean.py#L50) has a few examples of how to accomplish this.  

1. Define a function which returns a module to be tested (ie. [``test_mean_channel()``](https://github.com/NVIDIA-AI-IOT/torch2trt/blob/0f7a4594f5e796ae1f889e58936fa600d70dda3c/torch2trt/converters/mean.py#L53))

2. Register the module for testing by calling [``add_module_test``](https://github.com/NVIDIA-AI-IOT/torch2trt/blob/0f7a4594f5e796ae1f889e58936fa600d70dda3c/torch2trt/module_test.py#L41).

``add_module_test`` takes the following parameters

* ``dtype`` - the torch data type of inputs passed to the module
* ``device`` - the torch device of inputs passed to the module
* ``input_shapes`` - a list of tuples, specifying the shapes of inputs passed to the module for testing
* ``torch2trt_kwargs`` - any keyword arguments passed to ``torch2trt`` for the test (ie. ``fp16_mode = True``)

I think for now, it would be sufficient just to test ``dtype = torch.float32``, ``device = torch.device('cuda')``, and whichever input shapes are appropriate to test the module.

Once the module, for example ``test_convert_narrow`` is registered, you can test it by calling

```bash
python3 -m torch2trt.test --name=test_convert_narrow
```

The ``--name`` argument applies a regular expression to the full function path (including modules), so since it's defined in ``torch2trt.converters.slice`` you could also test by doing calling

```bash
python3 -m torch2trt.test --name=slice
```

Sorry we don't have contribution guidelines set up yet, I think that this will become more clear as we go through this process.  

Thanks again,
John

PR #28, opened 2019-07-26
added pad converter

jaybdub:


timegoby:
Input tensor is (B, C, H, W). The code can only pad H,W, but how can I pad C dimension?

PR #57, opened 2019-09-13
consider ceil_mode of torch.nn.MaxPool2d

mt1871:
If ceil_mode is False, the default value of layer.padding_mode is
PaddingMode.EXPLICIT_ROUND_DOWN. If ceil_mode is True, padding_mode
should be trt.PaddingMode.EXPLICIT_ROUND_UP.

### DIFF
--- a/torch2trt/converters/MaxPool2d.py
+++ b/torch2trt/converters/MaxPool2d.py
@@ -1,4 +1,5 @@
 from torch2trt.torch2trt import *
+from torch2trt.module_test import add_module_test
 
 
 @tensorrt_converter('torch.nn.MaxPool2d.forward')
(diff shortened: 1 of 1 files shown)


jaybdub:
Hi mt1871,

Thanks for sharing!  I think it would be good to add test cases for ``ceil_mode=True`` and ``ceil_mode=False`` before merging into master.

Please let me know if you can make these changes / have any questions, otherwise I'm happy to merge this into a separate branch and apply them.

Best,
John

mt1871:
Hi John,

I've just added the test cases. The test with name `torch2trt.converters.MaxPool2d.test_MaxPool2d_with_ceil_mode` would fail for input shape `[(1, 3, 4, 6)]` without passing the correct padding mode to TensorRT due to a high max error. Please let me know if this test is sufficient.

Best,
Markus

jaybdub:
Hi Markus,

I just verified the tests, and everything looks great.

Just merged into master.  Thanks for the contribution :) 

Best,
John

PR #58, opened 2019-09-13
added support for torch.add

mt1871:


mt1871:
PR withdrawn since feature has already been implemented in the meantime.

PR #76, opened 2019-09-30
add permute converter: issue#71

jeong-tae:
Hi, I am @jtlee who made a issue https://github.com/NVIDIA-AI-IOT/torch2trt/issues/71

Thank you for your help to make it work!


jeong-tae:
Permute is generalized function of transpose. so, If you want to combine transpose with permute, then control the dim arguments.

Still, I believe you/someone still want to separate function for permute and transpose. It will be good for management.

jeong-tae:
Seems already feature added

Issue #64, opened 2019-09-20
Doesn't support broadcasting


jtlee90:
For Add, generally numpy and pytorch allow to broadcast
For example, a tensor `A: shape (1, 3, 224, 224)` with other tensor `B: shape(1,)`,
`C = A + B`
is the tenser `C: shape (1, 3, 224, 224)`

you can easily reproduce error with

```
class Add(torch.nn.Module):
    def __init__(self):
        super(Add, self).__init__()
        self.constant = 1
    def forward(self, x):
        return x + self.constant
```

jtlee90:
And I found that whatever I change the `self.constant` to be same shape with `x`, still it doesn't have `_trt` attribute.
I made it to be same shape using expand_as function
`
def forward(self, x):
    self.constant = self.constant.expand_as(x).type(x.type())
    return x + self.constant
`
it returns not found '_trt' attribute error

PR #77, opened 2019-09-30
fix add to support broadcast

jeong-tae:
This PR is related to #64 (same person as jtlee who made issue)

I don't know which way is best for type casting and broad casting. but seems this is working.
Please consider broadcasting and typecasting case in add, which are I attached.

If you also think that casting require to implement, can you tell how can it possible(in smart way)?
and not only add method needs to broadcast but also mul, div, and so on. So, if this PR is approved, I'd like add more code on other method.

### DIFF
--- a/torch2trt/converters/add.py
+++ b/torch2trt/converters/add.py
@@ -9,6 +9,20 @@ def convert_add(ctx):
     input_a = ctx.method_args[0]
     input_b = ctx.method_args[1]
     output = ctx.method_return
+
+    if not hasattr(input_b, '_trt') and isinstance(input_b, torch.Tensor):
+        shape = tuple(input_a.shape[1:])
+        dtype_ones = torch.ones(shape, dtype = input_b.dtype).cpu().numpy()
+        values = dtype_ones * input_b.detach().cpu().numpy()
+        input_b._trt = ctx.network.add_constant(shape, values).get_output(0)
+    
+    if not hasattr(input_b, '_trt') and (isinstance(input_b, int) or isinstance(input_b, float)):
+        shape = tuple(input_a.shape[1:])
+        dtype_ones = torch.ones(shape, dtype = input_a.dtype).cpu().numpy()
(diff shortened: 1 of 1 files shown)


jaybdub:
Hi jeong-tae,

Thanks for sharing.  Just as a heads up, I've added ``trt_`` utility function in this branch

https://github.com/NVIDIA-AI-IOT/torch2trt/blob/0bfe27df7628b009ec2ebc990bc5ba2304ce1d7a/torch2trt/torch2trt.py#L102

It's usage is shown here

https://github.com/NVIDIA-AI-IOT/torch2trt/blob/0bfe27df7628b009ec2ebc990bc5ba2304ce1d7a/torch2trt/converters/normalize.py#L17

What it does is

1.  Add any missing trt tensors for leaf tensors / scalars
2.  Add shuffle layers to insert dimensions to make trt tensors broadcastable (if they're not already)
3.  Return set of broadcastable trt tensors

At least one torch.Tensor is required to infer the dtype/broadcast shape for scalar primitive types.  This should be okay since I don't think scalar primitives are used by themselves. 

This definitely needs many more test cases / usages, but the idea is to replace access to ``input._trt`` with for example ``trt_(input)``.  For example, for binary operations

```python
input_a = ctx.method_args[0]
input_b = ctx.method_args[1]
input_a_trt, input_b_trt = trt_(ctx.network, input_a, input_b)

# input_a_trt and input_b_trt are broadcastable
```

If you don't want to make tensors broadcastable, but do want to automatically add missing ``_trt`` for leaf tensors, I think you would just call ``trt_(...)`` separately on each tensor.

Of course, until this is merged it's highly subject to change.  

Please let me know if this helps you, or if you find any issues.

Best,
John

jtlee90:
cool! I saw your PR, which is well generalized so that it covers every method as well. nice.

jeong-tae:
I will close this PR, seems solved.

PR #79, opened 2019-10-01
fix half precision computation with normalize function

jeong-tae:
multiplication with cpu half precision not supported.
Need to specify where the device is if you use cuda().half()


jeong-tae:
#80 this can solve broadcast and type conversion problem in normalize as well
I will close.

PR #80, opened 2019-10-01
Prepare broadcastable tensors

jaybdub:


jeong-tae:
Hi, jaybdub!
This is very nice! Seems, now you can solve many issues!

I think adding test units for each converter would make code more nice.
Current test doesn't include broadcast test or type test.

Bests,
Jeongtae

jaybdub:
Hi jeong-tae, 

Thanks!  Currently the ``add_module_test`` method allows you to set the data type and shape of inputs.  It's mainly intended for cross-evaluating output, latency, and throughput against PyTorch under different configurations. 

https://github.com/NVIDIA-AI-IOT/torch2trt/blob/c0c89d2908b914c8350efb673ee076883d190093/torch2trt/module_test.py#L46

Do you mind sharing an example of what you mean by broadcast/type test?  

Best,
John

jtlee90:
oh, I am sorry. I didn't notice that you already added it in add_module_test
https://github.com/NVIDIA-AI-IOT/torch2trt/blob/master/torch2trt/converters/add.py#L57-L62
sorry for bothering you

Issue #82, opened 2019-10-02
Do we need to check is_leaf?


jtlee90:
https://github.com/NVIDIA-AI-IOT/torch2trt/blob/master/torch2trt/torch2trt.py#L127-L132

We employ TensorRT for accelerating inference in most of case, so I think checking the `is_leaf` may prevent converting 'weight' tensor. When we do the inference, we do not use a gradient(sometime for visualization used) and if tensor.requires_grad == True, then the tensor doesn't have `is_leaf` so that caused Assert error(if they use detach() it is okay) during the converting.

So, I mean, &*it is mainly for accelerating inference whatever the tensor is leaf or not, nobody will care**

I believe that these day a lot of research use self-attention mechanisms which use learnable weight parameter, like `input * weight`, then if with this code, it fail to generate trt_

What do you think? Do we have to check `is_leaf`, and what for?

jtlee90:
jaybdub's mention is here.
> FYI, The main reason we check is_leaf is to make sure we don't create a constant for a tensor that actually isn't constant. This can happen when there is a PyTorch operation that isn't yet supported/registered in torch2trt. When a layer isn't supported, the failure doesn't actually occur until we later try to access the _trt, which is missing.

My concern is that, if we always produce constants, this failure may not happen, and the conversion may succeed and silently produce an incorrect model.

jaybdub:
Hi jtlee90,

I've actually removed ``leaf_check`` as you suggested.  The methodology seemed not reliable when ``torch.no_grad`` was set.  This means that there can be silent failures (successful conversion, incorrect model).  However it seems``torch.jit`` has similar expectations

https://pytorch.org/docs/stable/jit.html#torch.jit.trace

Perhaps there is more that can be done to prevent this.  One simple (yet laborious) thing I can think of is manually list all PyTorch ops that modify tensors.  

```python
@tensorrt_converter('torch.unimplemented_op1')
@tensorrt_converter('torch.unimplemented_op2')
def convert_unimplemented(ctx):
    raise NotImplementedError
```

Another simple thing would be a sanity check at the end of conversion, that just cross-checks the final output against Python, and warns if a tolerance is exceeded.  It seems ``torch.jit`` also does this.

Please let me know if you have any suggestions.

Best,
John

PR #83, opened 2019-10-02
remove is_leaf checker

jtlee90:
I'd like to recommend removing `is_leaf` for learnable 'weight' tensor, which the tensor actually should be considered as leaf but it is not.
as mentioned in #82, `weight` can be used everywhere. For example, if you want to control `beta` mentioned in [retinaMask](https://arxiv.org/pdf/1901.03353.pdf) paper(see section 3.1: self-adjust smooth l1), it should possible to generate tensor._trt. Also, in this model [GAN](https://arxiv.org/pdf/1805.08318.pdf), they use learnable 'weight' parameter to control initial attention rate. Even Memory network, they abuse learnable weight parameter.

So, I hope this change not affect to model performance.

### DIFF
--- a/torch2trt/torch2trt.py
+++ b/torch2trt/torch2trt.py
@@ -124,7 +124,7 @@ def trt_(network, *tensors):
             trt_tensor = t._trt
             
         # or... add constant for leaf tensor w/o _trt
-        elif isinstance(t, torch.Tensor) and t.is_leaf and not hasattr(t, '_trt'):
+        elif isinstance(t, torch.Tensor) and not hasattr(t, '_trt'):
             # add leaf tensor
             shape = tuple(t.shape[1:])
             weight = t[0].detach().cpu().numpy()
@@ -136,7 +136,7 @@ def trt_(network, *tensors):
             shape = (1,) * broadcast_num_dim
             scalar = t * torch.ones(shape, dtype=dtype).cpu().numpy()
             trt_tensor = network.add_constant(shape, scalar).get_output(0)
-            
+         
         assert(trt_tensor is not None)#, 'TensorRT tensor could not be created')
             
         # MAKE TRT TENSOR BROADCASTABLE IF IT IS NOT ALREADY


jaybdub:
Hi jtlee90,

Thanks for sharing.   My current understanding is that PyTorch parameters are considered leaf tensors, though perhaps I'm missing an edge case.

FYI, The main reason we check ``is_leaf`` is to make sure we don't create a constant for a tensor that actually isn't constant.  This can happen when there is a PyTorch operation that isn't yet supported/registered in torch2trt.  When a layer isn't supported, the failure doesn't actually occur until 

1.  Another converter tries to access the _trt attribute but it's not there
2.  We mark the final network outputs by attempting to access the _trt attribute but it's not there

My concern is that, if we always produce constants, this failure may not happen, and the conversion may succeed and silently produce an incorrect model.

Is there a scenario you have in mind where a Parameter is not a leaf?  If there is something you're aware of, perhaps there is another solution.

Best,
John

jtlee90:
@jaybdub Thank you for your kind reply.

Well, It's true that PyTorch parameters are considered leaf tensors!
but you transform the parameters using function `unsqueeze` or `expand_as`, it is not leaf node anymore because it is not supported function?

What if I want to multiply or add parameter through input channels? For example, I have weight, shaped in 128, and inputs shaped in (1, 128, 64, 64). I have to use unsqueeze/expand_as to match size of tensor.
```
class WeightMul(torch.nn.Module):
    def __init__(self, size):
        super(WeightMul, self).__init__()
        self.w = torch.nn.Parameter(torch.Tensor(size))

    def forward(self, x):
        return self.w.unsqueeze(0).unsqueeze(2).unsqueeze(3).expand_as(x) * x


@add_module_test(torch.float32, torch.device('cuda'), [(1, 128, 64, 64)])
def test_rmul_weight():
    return WeightMul(128)
```
This test failed.

jaybdub:
Hi jtlee90,

I think the best solution to this problem may be to implement converters for ``unsqueeze`` and ``expand_as`` operations.  I believe this can be accomplished using the TensorRT shuffle layer.  

This will also fix other uses of ``unsqueeze`` and ``expand_as``.  

Apologies, many of the errors are opaque right now.  I've been considering registering ``dummy`` converters that throw verbose errors for layers that have not yet been implemented.  Of course this would be a "hint" since new API features may be introduced, or users create their own layers.  

Best,
John

jtlee90:
Okay. Thank you!
That's nice news. I am glad that there are many things to do! haha
I am trying to make this repository works in most of general cases so that anyone can utilize TensorRT with pytorch. It will accelerate pytorch that deep learning engineer use pytorch more, which make our research more easy. haha!

Anyway, I'd like to keep an eye on this reposity.

PR #157, opened 2019-11-15
nn.Upsample support with TRT IResizeLayer 

SrivastavaKshitij:
TensorRT 6.0.1.5 has a new [IResizeLayer](https://docs.nvidia.com/deeplearning/sdk/tensorrt-api/python_api/infer/Graph/Layers.html#iresizelayer) that performs interpolation. I have created a preliminary pull request to map this layer with `nn.Upsample`. 

**Note: This will only work with TRT 6 and above**. So @jaybdub do you think we need separate converter list based on TRT version in torch2trt going forward ?

Also, the [documentation](https://docs.nvidia.com/deeplearning/sdk/tensorrt-api/python_api/infer/Graph/Layers.html#iresizelayer) says that `kLinear` supports `linear`, `bilinear` and `trilinear`. However, I dont know how to specifically call `bilinear` and `trilinear` . So `linear` and `nearest` mode works fine as of now but we see some deviation in results for `bilinear` and `trilinear`


SrivastavaKshitij:
Also, I am not sure if my `torch.nn.functional.interpolate` implementation is correct. Could you please let me know what needs to be changed if any

jaybdub:
Hi SrivastavaKshitij,

Thanks for this pull request.  Here are two immediate things I would suggest

1. Creating just one file ``upsample.py``
2. Check the TensorRT version by adding a case statement _within_ the converter, so that this is used only if version supports it.  This will prevent breaking compatibility for previous versions.

There may be other edits and I will still need to test / validate the changes.

Please let me know if you have any questions.

Best,
John

SrivastavaKshitij:
Hi John 

1. I was taking cues from converter examples such as `relu.py` where I saw that `nn.functional.Relu` is  defined in a separate file that in turn called `RELU.py` where `nn.ReLU.forward` is defined. So I was following the same methodology. Do you suggest implementing `torch.nn.functional.interpolate` and `torch.nn.Upsample.forward` in the same file ?

I will get back to you with requested changes.

jaybdub:
Ah I see, sorry for the confusion.  I believe I did that with ReLU before I realized that the module ``torch.nn.ReLU`` actually calls ``torch.nn.functional.relu`` under the hood.  I probably should refactor this.

Actually, judging from 

https://pytorch.org/docs/1.0.0/_modules/torch/nn/modules/upsampling.html#Upsample

I believe it is true that ``torch.nn.Upsample.forward`` calls ``torch.nn.functional.interpolate`` internally.  Judging from the documentation, it looks like this is the case dating back to at least ``PyTorch 1.0`` 

Are you able to test if this is true?  

If this is true, I guess there are a couple methods.  

1.  Modify the converter for ``torch.nn.functional.interpolate`` to handle TensorRT 6

2.  Register a different converter depending on the TensorRT version

Best,
John

SrivastavaKshitij:
Yes, looking at the source code, `nn.Upsample` , `UpsamplingNearest2d` and `UpsamplingBilinear2d` call `F.interpolate` in the forward function. As you suggested, I will write a converter for interpolate and map all these nn layers on to that....

Secondly, could u help me figure out where in `torch2trt.py` file , converters are being registered so that I can go in and register converters based on trt version.

thanks

jaybdub:
Awesome, thanks!

You should not need to edit ``torch2trt.py``, but just FYI:

When you call [``@tensorrt_converter(...)``](https://github.com/NVIDIA-AI-IOT/torch2trt/blob/53ce0569132b277ccebaab371a007154caad6109/torch2trt/torch2trt.py#L387) it adds the converter to a global dict named [``CONVERTERS``](https://github.com/NVIDIA-AI-IOT/torch2trt/blob/master/torch2trt/torch2trt.py#L166).

I think in this instance, it may make sense to just handle the version checking inside the converter.  But either may be fine.  I primarily want to ensure that existing software will not break. 

Also just a heads up, it occurred to me that since ``bicubic`` interpolation is not supported in TensorRT 6, we will also need to fall back to the plugin in this instance.  

Best,
John

twmht:
@SrivastavaKshitij 

I got the following errors with this PR

```
[TensorRT] ERROR: (Unnamed Layer* 90) [Resize]: mismatch in number of dimensions for outputDims.
```

any idea?

SrivastavaKshitij:
Hi @twmht : So we have a bug open with nvidia regarding TRT IResizeLayer and once that bug is fixed , I will update this PR.

deepakanandece:
Was this done? Can I use nn.Upsample in bilinear mode now?

SrivastavaKshitij:
No. the bug with IResizeLayer will be fixed in trt version. So until then, there is no fix to this PR

VCBE123:
Now in TensorRT 7， the test of **interpolate layer** failed.

SrivastavaKshitij:
#249 is the superset PR of this PR that we should be monitoring for progress.

Issue #158, opened 2019-11-18
Interpolate Plugin [nn.Upsample vs F. Interpolate]


SrivastavaKshitij:
Hi John

While working on the pull request, I figured out that if I use interpolate plugin on `F.interpolate` it works fine otherwise there is an error when `nn.Upsample` is used. This happens for `bilinear` mode. I was able to reproduce the error:

1. Change [class Interpolate](https://github.com/NVIDIA-AI-IOT/torch2trt/blob/53ce0569132b277ccebaab371a007154caad6109/torch2trt/converters/interpolate/interpolate.py#L42) def  to:

```
class Interpolate(torch.nn.Module):
    def __init__(self, size, mode, align_corners,scale_factor=None):
        super(Interpolate, self).__init__()
        self.size = size
        self.mode = mode
        self.align_corners = align_corners
        self.scale_factor=scale_factor

    def forward(self, x):
        if self.scale_factor == None:
            return F.interpolate(x, self.size, mode=self.mode, align_corners=self.align_corners)
        else:
            return F.interpolate(x, scale_factor = self.scale_factor, mode = self.mode, align_corners = self.align_corners)

``` 

and add two unit tests at the end:
```
@add_module_test(torch.float32, torch.device('cuda'), [(1, 256, 192, 512)])
def test_scale():
    return Interpolate(None,'bilinear',False,scale_factor=2)

@add_module_test(torch.float32, torch.device('cuda'), [(1, 256, 192, 512)])
def test_nn_scale():
    return nn.Upsample(scale_factor = 2, mode='bilinear',align_corners=False)

```
I get the following results:

```
| torch2trt.converters.interpolate.interpolate.test_scale | float32 | [(1, 256, 192, 512)] | {} | 0.00E+00 | 613 | 785 | 1.66 | 1.33 |
| torch2trt.converters.interpolate.interpolate.test_nn_scale | float32 | [(1, 256, 192, 512)] | {} | 2.99E+00 | 605 | 914 | 1.68 | 1.15 |
```

jasonliu19:
I have noticed the same error, which is strange since I though nn.Upsample used interpolate under the hood.

SrivastavaKshitij:
Ok. I have a fix for the problem. will create a pull request tomorrow.

PR #160, opened 2019-11-19
fixed upsampling<->interpolate translation

SrivastavaKshitij:
Fixes issue #158 , `nn.Upsample` arguments were not correctly mapping to `F.interpolate`.


SrivastavaKshitij:
Issue was, arguments were not correctly being mapped. So in case of mode, `bilinear` would default to `nearest`, hence it would generate incorrect answers.

PR #168, opened 2019-11-26
Int8 support

jaybdub:
Adds support for int8 calibration

### Calibrate using inputs (Small data)

This shows basic usage where the calibration data is inferred from input data.  This will calibrate on 8 random normal samples.  

> Please note, the inference batch size remains whatever is set with ``max_batch_size`` parameter.

```python
data = torch.randn(8, 3, 224, 224).cuda()

model = torchvision.models.resnet18(pretrained=False).cuda().eval()

model_trt = torch2trt(model, [data], int8_mode=True)
```

### Calibrate using dataset (Big data)

Sometimes, the calibration data may be larger than can fit in memory.  In this case, you can provide a dataset class that implements the ``__len__`` and ``__getitem__`` methods.   Below we will implement a calibration dataset that functions similar to Example 1, but using the dataset method.

First, define the dataset

```python
class RandomDataset(object):

    def __len__(self):
        return 8

    def __getitem__(self, idx):
        return [ torch.randn(3, 224, 224) ]
```

Internally, the shape, device, and data type of calibration data will be inferred from the ``inputs`` parameter.  So it is OK if the tensors returned by ``__getitem__`` live on CPU/GPU or include/exclude the batch dimension.  It is important that the items returned by ``__getitem__`` have the correct shape and order of inputs in the case of multiple input nodes.

```python
dataset = RandomDataset()

model_trt = torch2trt(model, [data], int8_mode=True, int8_calib_dataset=dataset)
```

### Calibration algorithm

You may set the calibration algorithm, [see this](https://docs.nvidia.com/deeplearning/sdk/tensorrt-api/python_api/infer/Int8/Calibrator.html).

```python
model_trt = torch2trt(model, [data], int8_mode=True, int8_calib_algorithm=trt.CalibrationAlgoType.ENTROPY_CALIBRATION_2)
```

### DIFF
--- a/torch2trt/calibration.py
+++ b/torch2trt/calibration.py
@@ -16,20 +16,17 @@ class TensorBatchDataset():
     
 class DatasetCalibrator(trt.IInt8Calibrator):
     
-    def __init__(self, dataset, batch_size=1, algorithm=trt.CalibrationAlgoType.ENTROPY_CALIBRATION_2):
+    def __init__(self, inputs, dataset, batch_size=1, algorithm=trt.CalibrationAlgoType.ENTROPY_CALIBRATION_2):
         super().__init__()
         
         self.dataset = dataset
         self.batch_size = batch_size
         self.algorithm = algorithm
         
-        # pull sample, should not include batch dimension
-        inputs = dataset[0] 
-        
         # create buffers that will hold data batches
         self.buffers = []
         for tensor in inputs:
-            size = (batch_size,) + tuple(tensor.shape)
+            size = (batch_size,) + tuple(tensor.shape[1:])
             buf = torch.zeros(size=size, dtype=tensor.dtype, device=tensor.device).contiguous()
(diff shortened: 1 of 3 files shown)


torch2trt/calibration.py
```diff
@@ -0,0 +1,69 @@
...
+    def get_batch(self, *args, **kwargs):
+        if self.count < len(self.dataset):
+            
+            for i in range(self.batch_size):
+                
+                idx = self.count % len(self.dataset) # roll around if not multiple of dataset
+                inputs = self.dataset[idx]
+                
+                # copy data for (input_idx, dataset_idx) into buffer
+                for j, tensor in enumerate(inputs):
```
alsrgv: This fails if dataset returns (input, label) pairs.

How about this?
```python
for buffer, tensor in zip(self.buffers, inputs):
  buffer[i].copy_(tensor)
```

PR #175, opened 2019-12-04
implemented support for tensorrt_method

jaybdub:


jaybdub:
This PR introduces a new API for flagging custom layers for TensorRT conversion.  It is the inverse of ``tensorrt_converter`` with the added benefit of not requiring global name resolution.

## Usage

To attach a converter to a custom method, use the decorator``@tensorrt_method``, passing the converter as an argument.

```python

def convert_my_layer(ctx):
    # code to convert my_layer (see torch2trt README)

@tensorrt_method(convert_my_layer)
def my_layer(x):
    # some custom layer...
```

Then you can convert a module using the layer

```python

class Model(nn.Module):

    def forward(self, x):
        x = torch.relu(x)
        x = my_layer(x)
        return x

model = Model().cuda().eval()

data = torch.randn(1, 3, 224, 224).cuda()

model_trt = torch2trt(model, [data])
```

Class methods can be treated similarily

```python
def convert_MyModule(ctx):
    # code to convert...

class MyModule(nn.Module):

    @tensorrt_method(convert_MyModule)
    def forward(self, x):
        # some custom code
```

PR #292, opened 2020-04-09
Support passing padding mode in ConvTranspose2d

salexspb:
This is a dirty hack for now - would be great to hear suggestions to do it properly. Here is what his hack allows me to do:

```
        model = torch.nn.ConvTranspose2d(
            64,
            64,
            3,
            stride=(1, 2),
            padding=(1, 1),
            output_padding=(0, 1),
            bias=False,
        )
        import tensorrt
        model.trt_padding_mode = tensorrt.PaddingMode.SAME_LOWER
        _test_trt_convertion(model, make_input([1, 64, 64, 1792]))  # just a function which converts a model using torch2trt and validates that results didn't change
```

Without setting the mode, I can not get TRT's Deconvolution to produce output dimension which is exactly 2x larger than the original one (i.e. upsample 2x via ConvTranspose2d in pytorch). Pytorch can handle this via output_padding parameter.

This seems to be luck that it works right now. What would be a proper solution?

I see this code in ONNX converter, but it is quite involved: https://github.com/onnx/onnx-tensorrt/blob/master/onnx2trt_utils.cpp#L621 Is there any python API analogue for this?


salexspb:
cc @jaybdub

Issue #311, opened 2020-05-08
ConvTranspose1d converter


tambien:
It seems like it's possible to create torch.nn.ConvTranspose1d by reshaping the input and output of the torch.nn.ConvTranspose2d converter. I don't have any experience writing converters. Could someone point me in the right direction of how you would do that for this layer? 

Thanks for any help or guidance!

jaybdub:
Hi tambien,

Thanks for reaching out.  We do this for Conv1d, this file might help guide you

https://github.com/NVIDIA-AI-IOT/torch2trt/blob/master/torch2trt/converters/Conv1d.py#L5

Please let me know if you have any success with this.

Best,
John

tambien:
🙏🏻I'll give this a shot! if it works out, i'll try contributing it back

PR #314, opened 2020-05-12
Converter for ConvTranspose1d

tambien:
addresses NVIDIA-AI-IOT/torch2trt#311

### DIFF
--- a/torch2trt/converters/ConvTranspose1d.py
+++ b/torch2trt/converters/ConvTranspose1d.py
@@ -1,0 +1,43 @@
+from torch2trt.torch2trt import *
+from torch2trt.module_test import add_module_test
+
+
+@tensorrt_converter('torch.nn.ConvTranspose1d.forward')
+def convert_ConvTranspose1d(ctx):
+    module = ctx.method_args[0]
+    input = ctx.method_args[1]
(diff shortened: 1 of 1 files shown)


jaybdub:
Hi tambien,

Thanks for this pull request!

Before merging the PR, we would like to add some test cases to ensure the intended use cases work.  Please see this file for an example

https://github.com/NVIDIA-AI-IOT/torch2trt/blob/master/torch2trt/converters/Conv1d.py#L47

Also, this is not required, but I'm curious what application the ConvTranspose1d layer is used in.  If you're able to, do you mind sharing the model / application that the ConvTranspose1d is used in?  

Best,
John

tambien:
just pushed a couple of tests grabbed from Conv1d. let me know if that's sufficient!

i need this layer for some audio processing that i'm playing around with. nothing public/published yet, just trying to get things to work. this library has been really helpful in speeding things up!

jaybdub:
Thanks!  I'll test it out.

Ah, audio processing, cool!  This is an area less tested with this repository.  Glad to hear it's working. 

Please let me know if you make further progress, or think of other features for torch2trt that would help in this domain.

Best,
John

PR #328, opened 2020-06-04
ConvTranspose2d with unit test

SrivastavaKshitij:
**This PR supersedes #321 , fixed indentation error in #321 and adds unit tests**

I changed the default workspace size in `torch2trt.py` because for some reason I was getting the following error with TransposeConv2d unit tests:

```[TensorRT] ERROR: Internal error: could not find any implementation for node (Unnamed Layer* 202) [Deconvolution], try increasing the workspace size with IBuilder::setMaxWorkspaceSize()```

Changing the workspace size has no effect on other ops.


jaybdub:
Added test case for kernel_size 4 and merged.

Thank you!

Issue #216, opened 2019-12-26
Hello, this is very convenient repo to convert pytorch model to tensorrt, but it likely support cv well, support nlp poolly,I try to convert gpt2 model to tensorRT, but there a


bigprince97:

bigprince97:
Hello, this is very convenient repo to convert pytorch model to tensorrt, but it likely support cv well, support nlp poolly,I try to convert gpt2 model to tensorRT, but there are two operations that transformers(nlp very popular model, it is the base block to build gpt2, bert) needs don't support, include matmul, layernorm, c++ api support it, consider support it in python api? I am very expected to deploy gpt2 in tensorRT.

bigprince97:
I convert gpt2 model to tensorRT successfully use this repo and some operation I write by myself, and the speed has been increased three times, thanks this good repo.

mowayao:
> I convert gpt2 model to tensorRT successfully use this repo and some operation I write by myself, and the speed has been increased three times, thanks this good repo.

Can you share you impl of matmul, plz?

czs1886:
> > I convert gpt2 model to tensorRT successfully use this repo and some operation I write by myself, and the speed has been increased three times, thanks this good repo.
> 
> Can you share you impl of matmul, plz?

Just use  addMatrixMultiply() which is defined [here](https://docs.nvidia.com/deeplearning/tensorrt/api/c_api/classnvinfer1_1_1_i_network_definition.html#a6197f3576f446c22c3b60ef60e9e8c62)
You need to create a new converter in converters directory. You can copy mul.py and change its addElementwise to addMatrixMultiply.

q248953144:
> I convert gpt2 model to tensorRT successfully use this repo and some operation I write by myself, and the speed has been increased three times, thanks this good repo.

can you share your ways

Fan9:
@bigprince97 can you  shart your code? thanks!

bigprince97:
Sorry to see the reply now， I write code last year, it can't run because of the update of the repo. fastertransformer supports gpt2 acceleration，It has pytorch interface， you can try.

francesco-taioli:
for anyone looking for the ```matmul```, here's the code
```
@tensorrt_converter('torch.Tensor.__matmul__')
def convert_mul(ctx):
    input_a = ctx.method_args[0]
    input_b = ctx.method_args[1]
    output = ctx.method_return
[...]

PR #334, opened 2020-06-10
matmul

czs1886:
addresses #216 
This could solve the matmul unknown issue.


liaopeiyuan (approved):
This works perfectly for me!

PR #357, opened 2020-07-02
Update requirements

vellamike:


jaybdub:
Hi vellamike,

Thanks for the PR.  The installation instructions vary a bit from platform to platform from what is set here (for example, on Jetson torch is installed from [pre-built wheel](https://forums.developer.nvidia.com/t/pytorch-for-jetson-nano-version-1-5-0-now-available/72048/2)).  Some use docker containers with specific versions of cuda/torch installed.

That said, it would be good to make sure developers can get up and running however possible.

Could you clarify which platform (ie:  GPU) you're using?  What was the biggest pain point in getting up and running on your platform?

Best,
John

vellamike:
Hi @jaybdub  - I'm running on a GV100. It was actually quite simple for me to get running, the main issues were:

1. Having TRT installed and in my library path
2. Having the TRT python API installed
3. The missing python libraries.

Maybe I could change the installation instructions to clarify that these are for non-Jetson devices? We could have different installation instructions depending on environment?

jaybdub:
Hmm I see.  For those new to TensorRT, I can see how this may be a pain point.

Would a note at the top like this have helped? 

---

## Setup

> torch2trt depends on the TensorRT Python API.  On Jetson, this is included with the latest JetPack.  For desktop, please follow the [TensorRT Installation Guide](https://docs.nvidia.com/deeplearning/tensorrt/install-guide/index.html).  You may also try installing torch2trt inside one of the NGC PyTorch docker containers for [Desktop](https://ngc.nvidia.com/catalog/containers/nvidia:pytorch) or [Jetson](https://ngc.nvidia.com/catalog/containers/nvidia:l4t-pytorch).

### Option 1 - Without plugins

To install without compiling plugins, call the following

```bash
git clone https://github.com/NVIDIA-AI-IOT/torch2trt
cd torch2trt
sudo python setup.py install
```

### Option 2 - With plugins (experimental)

To install with plugins to support some operations in PyTorch that are not natviely supported with TensorRT, call the following

> Please note, this currently only includes the interpolate plugin.  This plugin requires PyTorch 1.3+ for serialization.  

```bash
sudo python setup.py install --plugins
```

---

For now i'd like to avoid strictly enforcing dependency installation, since I'm not sure if there are adverse side effects.  For non-compiled libraries, and without adding strict (==) checks, it is probably OK though

vellamike:
> Would a note at the top like this have helped?

Yes I think this makes things more clear.

Issue #329, opened 2020-06-05
AttributeError: module 'tensorrt' has no attribute '__version__'


danial880:
I am using torch2trt on jetson nano. I have installed torch2trt with plugins using the official document. I am using the latest jetpack and version(7.1.0.16) of tensorrt. I get the following error

import torch2trt
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/dev/.virtualenvs/pytorch/lib/python3.6/site-packages/torch2trt/__init__.py", line 1, in <module>
    from .torch2trt import *
  File "/home/dev/.virtualenvs/pytorch/lib/python3.6/site-packages/torch2trt/torch2trt.py", line 5, in <module>
    from .calibration import TensorBatchDataset, DatasetCalibrator, DEFAULT_CALIBRATION_ALGORITHM
  File "/home/dev/.virtualenvs/pytorch/lib/python3.6/site-packages/torch2trt/calibration.py", line 5, in <module>
    if trt.__version__ >= '5.1':
AttributeError: module 'tensorrt' has no attribute '__version__'

LearnedVector:
same error here on an ubuntu machine.

lifefeel:
You may have `tensorrt.py` file on current directory. Delete this first and re-run.

AmitNativ1984:
I had the same issue. Running on Jetpack 4.6.2. Had to update Tensorrt on host to 8.2.1.9 from 8.2.1.8.
* Here is a link to the patch from Nvidia: https://developer.nvidia.com/embedded/linux-tegra-r3272.
* Download and extract the [tar.gz](https://developer.nvidia.com/tensorrt-8219-patch-jetpack46-jetson-nx-16gbtargz).
* Follow the instructions in the install.md readme file to install it.
* Also make sure you have the Tensorrt csv file on host at: /etc/nvidia-container-runtime/host-files-container.d/tensorrt.csv
* If you don't you can add it with: 
`sudo apt-get install nvidia-container-csv-tensorrt`

CNLiuZihao:
> You may have `tensorrt.py` file on current directory. Delete this first and re-run.

thanks

PR #390, opened 2020-08-25
Torch repeat [Problem: Adds an extra dimension after the conversion]

SrivastavaKshitij:
### Extra file

I have added an extra file in the base directory `repeat_test.py` but an extra dimension is being added during conversion. 

Run command:

```
cd torch2trt
python repeat_test.py
```

Output: 

```
root@388803c8185c:/workspace/work/torch2trt# python repeat_test.py 
[TensorRT] ERROR: INVALID_ARGUMENT: Cannot find binding of given name: input_0
torch.Size([10])
tensor([1., 1., 1., 1., 1., 1., 1., 1., 1., 1.], device='cuda:0')
torch.Size([1, 10])
tensor([[1., 1., 1., 1., 1., 1., 1., 1., 1., 1.]], device='cuda:0')

```

As per Yu-Te's guidelines , I changed 

```
https://github.com/NVIDIA-AI-IOT/torch2trt/blob/master/torch2trt/torch2trt.py#L129 to num_dim = len(t.shape[1:])
and
https://github.com/NVIDIA-AI-IOT/torch2trt/blob/master/torch2trt/torch2trt.py#L149 to shape = tuple(t.shape[1:])
```

but it doesnt solve the problem. I havent pushed the above mentioned changes on this branch, I tried those changes locally .

### DIFF
--- a/torch2trt/converters/repeat.py
+++ b/torch2trt/converters/repeat.py
@@ -1,3 +1,4 @@
+import tensorrt as trt 
 from torch2trt.torch2trt import *
 from torch2trt.module_test import add_module_test
 
@@ -5,25 +6,23 @@ from torch2trt.module_test import add_module_test
 def convert_repeat(ctx):
     inputs = get_arg(ctx, 'input', pos=0, default=None) 
     output = ctx.method_return
-    op_shape = output.size().as_list()
-    in_shape =inputs.size().as_list()
-    print(op_shape,in_shape)
-    layer = ctx.network.add_concatenation(inputs=trt_inputs)
+    output_list = output.detach().cpu().numpy()
+    op_shape = list(output.size())
+    layer = ctx.network.add_constant(shape=trt.Dims(op_shape),weights=trt.Weights(output_list))
     output._trt = layer.get_output(0)
 
 class Repeat(torch.nn.Module):
     def __init__(self, arg):
         super(Repeat, self).__init__()
         self.arg = arg
-        print("self.arg",self.arg)
 
     def forward(self, x):
         return x.repeat(self.arg)
 
-@add_module_test(torch.float32, torch.device('cuda'), [(1, 4, 4)] , enabled=trt_version() >= '6.0')
+@add_module_test(torch.float32, torch.device('cuda'), [(4, 4)] , enabled=trt_version() >= '6.0')
 def test_Stack_repeat():
-    return Repeat(3)
+    return Repeat(3,3)
 
-@add_module_test(torch.float32, torch.device('cuda'), [(1, 4, 4)], enabled=trt_version() >= '6.0')
+@add_module_test(torch.float32, torch.device('cuda'), [(4)], enabled=trt_version() >= '6.0')
 def test_basic2_repeat():
-    return Repeat(1)
+    return Repeat(10)
--- a/torch2trt/converters/repeat.py
+++ b/torch2trt/converters/repeat.py
@@ -1,0 +1,29 @@
+from torch2trt.torch2trt import *
+from torch2trt.module_test import add_module_test
+
+@tensorrt_converter('torch.Tensor.repeat', enabled=trt_version() >= '6.0')
+def convert_repeat(ctx):
+    inputs = get_arg(ctx, 'input', pos=0, default=None) 
+    output = ctx.method_return
+    op_shape = output.size().as_list()
+    in_shape =inputs.size().as_list()
+    print(op_shape,in_shape)
+    layer = ctx.network.add_concatenation(inputs=trt_inputs)
+    output._trt = layer.get_output(0)
+
+class Repeat(torch.nn.Module):
+    def __init__(self, arg):
+        super(Repeat, self).__init__()
+        self.arg = arg
+        print("self.arg",self.arg)
+
+    def forward(self, x):
+        return x.repeat(self.arg)
+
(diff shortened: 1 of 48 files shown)


torch2trt/converters/repeat.py
```diff
@@ -0,0 +1,28 @@
...
+from torch2trt.torch2trt import *
+from torch2trt.module_test import add_module_test
+
+@tensorrt_converter('torch.Tensor.repeat', enabled=trt_version() >= '6.0')
+def convert_repeat(ctx):
+    inputs = get_arg(ctx, 'input', pos=0, default=None) 
+    output = ctx.method_return
+    output_list = output.detach().cpu().numpy()
+    op_shape = list(output.size())
+    layer = ctx.network.add_constant(shape=trt.Dims(op_shape),weights=trt.Weights(output_list))
```
jaybdub, line 11: Looks like this is adding a constant, which may give some false-positive conversions (where a dynamic input is added as a constant).  

Have you looked into using different TensorRT method to supporting non-constant inputs?

SrivastavaKshitij: I tried `add_concatenation` method. However, it only concatenates a list of input tensors and I can't explicitly say on which dimension to concatenate. So it will fail on the following use case:

```
x = torch.tensor([1, 2, 3])
x.repeat(4, 2)
```

I cant think of any other method.

jaybdub: Hmm, I do see that it doesn't allow axis specification for ``add_concatenation``

https://docs.nvidia.com/deeplearning/tensorrt/api/python_api/infer/Graph/Network.html#tensorrt.INetworkDefinition.add_concatenation

I haven't personally tried, but it looks like there is an ``axis`` variable for the returned layer which controls this.

https://docs.nvidia.com/deeplearning/tensorrt/api/python_api/infer/Graph/Layers.html#iconcatenationlayer

Have you tried setting this variable after adding the layer?

```python
layer = network.add_concatenation(...)
layer.axis = ...
```

SrivastavaKshitij:
Oh yes, that's correct. I will try it tomorrow and get back to you.

Thanks

jaybdub:
No problem, let me know how it goes.  

Also, I'm not sure how it compares in terms of performance, but another option may be using addition by a zero tensor of some desired shape, and relying on broadcasting to repeat the input tensor.  I'm not sure if this breaks in certain cases though.

SrivastavaKshitij:
@jaybdub : It doesnt work. 

Apparently, IConcatenate Layer expects `NCHW / NHWC` format tensors to perform concatenation. Any tensor smaller than 4 dimension, it doesnt allow me to change the axis. 

```
concat axis must be less than number of non-batch dimensions (number of dimensions = 0, axis = 1).
```

If you run `python repeat_test.py` under torch2trt folder, you will see the error.

jaybdub:
Hi @SrivastavaKshitij ,

Sorry for the delayed response / activity here.

Let me look into it.  Just to clarify, the "constant" bug, you're referring to the dimension mis-match when constants are inserted, correct?

Best,
John

SrivastavaKshitij:
Yes, John . Please use this commit number: 822f413 . on this PR

I kept commit number : c2406b0 just to reference that concat will not work. 

Thanks.

jaybdub:
Hi @SrivastavaKshitij ,

I believe I've addressed the issue regarding the shape mismatch for constants that include a batch dimension.

Please try this pull-request (it is based off master).

https://github.com/NVIDIA-AI-IOT/torch2trt/pull/397

Please let me know if this resolves the issue you were facing, or you run into any issues.

Best,
John

SrivastavaKshitij:
Hi @jaybdub :

For the following script , PR #397 solves the problem 👍 

```
import torch

class ImageNorm(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.mean = torch.Tensor([0.485,0.456,0.406]).view(1,3,1,1).cuda().float()
        self.std = torch.Tensor([0.229,0.224,0.225]).view(1,3,1,1).cuda().float()

    def forward(self,inputs):
        inputs = torch.sub(inputs,self.mean)
        inputs = torch.div(inputs,self.std)
        return inputs

model= ImageNorm().cuda().eval()
inputs = [torch.randn((1,3,4,7),dtype=torch.float32,device="cuda")]
print(model(*inputs).size())

model_trt = torch2trt(model, inputs, max_batch_size=1, input_names=['inp'], output_names=['out'])
print(model_trt(*inputs).size())

```

However, for repeat converter I get an error first and then see the extra dimension:

Command:
```
cd torch2trt/
python repeat_test.py
```

Output

```
[TensorRT] ERROR: INVALID_ARGUMENT: Cannot find binding of given name: in
torch.Size([3, 3])
torch.Size([1, 3, 3])
```

I think if `invalid argument` can be solved then we wont see the extra dimension. Any ideas on how to fix this ?

Thanks so much

Kshitij

jaybdub:
I noticed in your ``repeat_test.p`` for the commit hash you shared that you are providing an input with size ``torch.ones([1])``.  

TensorRT assumes the first dimension of inputs is the batch size, I haven't yet thought if/how this could be handled.

Does your use case require ``torch.ones([1])`` instead of say ``torch.ones([1, 1])``?

SrivastavaKshitij:
Hi @jaybdub : 

I tried it and it doesnt work. We can leave `torch.repeat` for now. 

John, could you please merge PR #329 on master as it solves a major problem . 

PS: I will visit `repeat converter` later on. 

Thanks
Kshitij Srivastava

PR #393, opened 2020-08-28
Narrow converter

SrivastavaKshitij:
Added a converter for torch.narrow

Would like to thank @jinzhongxiao for his PR #23 which helped me unblock here.


jaybdub:
Looks great, just ran your test cases.

 Thanks @SrivastavaKshitij !

PR #394, opened 2020-08-31
Fix gpu memory leak

KibaAmor:
Memory leak can be reproduced by code:

```python3
import torch
from torchvision.models.resnet import resnet18

model = resnet18(pretrained=False).eval().cuda()

for i in range(100):
    x = torch.ones((1, 3, 224, 224)).cuda()
    model_trt = torch2trt(model, [x])
    y = model_trt(x)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ model = alexnet(pretrained=True).eval().cuda()
 x = torch.ones((1, 3, 224, 224)).cuda()
 
 # convert to TensorRT feeding sample data as input
-model_trt = torch2trt(model, [x])
+with torch2trt(model, [x]) as model_trt:
 ```
 
 ### Execute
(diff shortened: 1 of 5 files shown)


jaybdub:
Hi @KibaAmor ,

I will try to test your code soon.

Thanks for pointing this out and making this pull request.  My understanding was that the garbage collector handled the destruction of contexts/engines.

Just curious, have you tested manually triggering a garbage collection to see if this is the case? See [This package](https://docs.python.org/3/library/gc.html) for ``gc.collect()``.

The context managers may still be useful to give users explicit control of this, I'm just curious to better understand the circumstances this occurs.

Best,
John

KibaAmor:
I try with code:
```python3
import gc
import time

import torch
from torchvision.models.resnet import resnet18

model = resnet18(pretrained=False).eval().cuda()


def infer():
    x = torch.ones((1, 3, 224, 224)).cuda()
    model_trt = torch2trt(model, [x])
    y = model_trt(x)


for i in range(100):
    t1 = time.time()
    infer()
    t2 = time.time()
    torch.cuda.empty_cache()
    t3 = time.time()
    gc.collect()
    t4 = time.time()
    print(f'iter: {i}, infer: {t2 - t1}, empty_cached: {t3 - t2}, gc: {t4 - t3}')
```

but I still got out of memory, Any misunderstanding with me for your means?

PR #396, opened 2020-08-31
softmax neg dim

jaybdub:


JackieZhai:
what if "dim=0" ?
it seems to be error at "1 << (dim - 1)" !

PR #398, opened 2020-09-02
expose the batch size of INT8 calibration as parameter

Chujingjun:
Different batch size may generate different quantization result.
In my case, larger batch size increase the accuracy of the quantized model.


jaybdub:
Looks great, thank you @Chujingjun !

PR #430, opened 2020-10-19
QAT workflow with FP16 fallback support

SrivastavaKshitij:
This workflow is based on `Pytorch>=1.6` Quantize Aware Training library

**Note** : I have added a working example under `torch2trt/qat_workflow`. This directory is just for demo purposes and will be restructured / removed when the PR is finally merged. 

To use the qat_workflow, run the following command

```
cd torch2trt/qat_workflow
python infer.py
```

### Highlights of the PR

1. This PR uses `torch.nn` library for non quantized version and `torch.nn.intrinsic.qat` for quantized version of the model. Please refer `torch2trt/qat_workflow/utilities.py` 
2. Added an `if` condition under `converters/__init__.py` file to make sure that QAT ops are registered only for `pytorch>=1.6`
3. Explicitly added fallback support for conv, bn and relu layers to fallback to FP16 instead of FP32. 
4. Removed calibration process for QAT as it is not required. `Line 534` in `torch2trt.py`

### To Do
1. Add converter for QAT Linear layer.


jaybdub:
Hi @SrivastavaKshitij ,

Thanks for this PR!  

There is a good amount to review here, so it may take some time to get through, but I hope to enable this.

Please let me know of any changes / progress to the state of this PR, or feel free to request a review as needed.

Best,
John

SrivastavaKshitij:
@jaybdub : This is Pytorch workflow. I am creating another PR with Nvidia quantization library. I would like you to look at that PR first. 

THanks

PR #434, opened 2020-10-27
New Converter and Plugin for GroupNorm - Enables fallback to pytorch torch.nn.group_norm

sandeepkumar-skb:
Adding a plugin and converter for `GroupNorm` in Torch2trt. 
Since `GroupNorm` is not implemented in TensorRT yet this will allow a mechanism to fallback to Pytorch when it is called from the model. 

This PR has the following changes:
1. `torch2trt/plugins/group_norm.cpp` 
    - This is the plugin that launches the pytorch `GroupNorm` op - `at::group_norm()`
2. `torch2trt/converters/group_norm.py` 
    - Converter which gets invoked when `torch.nn.group_norm` is used in the users network 
    - This converter will call the plugin to do the computation on the tensor and returns the value.
3. `torch2trt/plugins/plugins.cpp`
    - Refactored the structure of `torch2trt/plugins` 
    - Added a new file `torch2trt/plugins/plugins.cpp`: This is the master file where all the plugins are registered and this will be provided to setup.py to build all the plugins.

**Environment Settings:** 
(Tested on 20.03 and 20.06 NGC containers)
PyTorch Versions: 1.5 and 1.6 
TensorRT Versions: 7.0 and 7.1
Platform: TitanRTX

**Attaching the results of:**
`python3 -m torch2trt.test --name=converters --tolerance=1e-2`
[group_norm_test.txt](https://github.com/NVIDIA-AI-IOT/torch2trt/files/5442200/group_norm_test.txt)

**To run only the group_norm converter:**
`python -m torch2trt.test --name=group_norm`

@jaybdub

### DIFF
--- a/build.py
+++ b/build.py
@@ -5,6 +5,7 @@ from string import Template
 
 PLUGINS = [
     'interpolate',
+    'group_norm',
 ]
 
 BASE_FOLDER = 'torch2trt/converters'
--- a/setup.py
+++ b/setup.py
@@ -11,11 +11,10 @@ def trt_lib_dir():
 
 ext_modules = []
 
-plugins_ext_module = [
-    CUDAExtension(
-        name='interpolate_plugin', 
+plugins_ext_module = CUDAExtension(
+        name='plugins', 
         sources=[
-            'torch2trt/plugins/interpolate.cpp'
+            'torch2trt/plugins/plugins.cpp'
         ],
         include_dirs=[
             trt_inc_dir()
@@ -30,30 +29,9 @@ plugins_ext_module = [
             'cxx': ['-DUSE_DEPRECATED_INTLIST'] if torch.__version__ < "1.5" else [],
             'nvcc': []
         }
-    ),
-    CUDAExtension(
-        name='group_norm_plugin', 
-        sources=[
-            'torch2trt/plugins/group_norm.cpp'
-            ],
-        include_dirs=[
-            trt_inc_dir()
-            ],
-        library_dirs=[
-            trt_lib_dir()
-            ],
-        libraries=[
-            'nvinfer'
-            ],
-        extra_compile_args={
-            'cxx': ['-DUSE_DEPRECATED_INTLIST'] if torch.__version__ < "1.5" else [],
-            'nvcc': []
-            }
-        ),
-]
-
+    )
 if '--plugins' in sys.argv:
-    ext_modules.extend(plugins_ext_module)
+    ext_modules.append(plugins_ext_module)
     sys.argv.remove('--plugins')
     
 
--- a/setup.py
+++ b/setup.py
@@ -11,8 +11,9 @@ def trt_lib_dir():
 
 ext_modules = []
 
-plugins_ext_module = CUDAExtension(
-        name='plugins', 
+plugins_ext_module = [
+    CUDAExtension(
+        name='interpolate_plugin', 
         sources=[
             'torch2trt/plugins/interpolate.cpp'
         ],
@@ -29,10 +30,30 @@ plugins_ext_module = CUDAExtension(
             'cxx': ['-DUSE_DEPRECATED_INTLIST'] if torch.__version__ < "1.5" else [],
             'nvcc': []
         }
-)
+    ),
+    CUDAExtension(
+        name='group_norm_plugin', 
+        sources=[
+            'torch2trt/plugins/group_norm.cpp'
+            ],
+        include_dirs=[
+            trt_inc_dir()
+            ],
+        library_dirs=[
+            trt_lib_dir()
+            ],
+        libraries=[
+            'nvinfer'
+            ],
+        extra_compile_args={
+            'cxx': ['-DUSE_DEPRECATED_INTLIST'] if torch.__version__ < "1.5" else [],
+            'nvcc': []
+            }
+        ),
+]
 
 if '--plugins' in sys.argv:
-    ext_modules.append(plugins_ext_module)
+    ext_modules.extend(plugins_ext_module)
     sys.argv.remove('--plugins')
     
 
--- a/torch2trt/converters/__init__.py
+++ b/torch2trt/converters/__init__.py
@@ -32,6 +32,7 @@ from .getitem import *
 from .identity import *
 from .instance_norm import *
 from .interpolate import *
+from .group_norm import *
 from .max import *
 from .max_pool2d import *
 from .mean import *
--- a/torch2trt/converters/group_norm.py
+++ b/torch2trt/converters/group_norm.py
@@ -39,7 +39,7 @@ def test_group_norm_trt_g2_fp32():
     return torch.nn.GroupNorm(2, 10)
 
 @add_module_test(torch.float32, torch.device('cuda'), [(1, 10, 112, 112)], has_group_norm_plugin())
-def test_group_norm_trt_g2_fp16():
+def test_group_norm_trt_g2_eps_fp32():
     return torch.nn.GroupNorm(2, 10, eps=1e-4)
 
 
--- a/torch2trt/converters/group_norm.py
+++ b/torch2trt/converters/group_norm.py
@@ -5,16 +5,15 @@ import collections
 
 def has_group_norm_plugin():
     try:
-        #from torch2trt.plugins import GroupNormPlugin
-        from torch2trt.group_norm_plugin import GroupNormPlugin
+        from torch2trt.plugins import GroupNormPlugin
         return True
     except:
         return False
 
 
 def get_group_norm_plugin(num_groups): #, num_channels, height_width, eps):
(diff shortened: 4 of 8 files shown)


jaybdub:
Hi @sandeepkumar-skb ,

Thanks for the great PR!

I just tested on my system, and I noticed the PyTorch cross-validation error is quite large (>4 max error).

```
| torch2trt.converters.group_norm.test_group_norm_trt_g2_fp32 | float32 | [(1, 10, 112, 112)] | {} | 4.44E+00 | 7.57e+03 | 3.29e+03 | 0.18 | 0.301 |
| torch2trt.converters.group_norm.test_group_norm_trt_g2_eps_fp32 | float32 | [(1, 10, 112, 112)] | {} | 4.18E+00 | 7.45e+03 | 5.01e+03 | 0.18 | 0.294 |
```

I noticed a similar output in your test case.  

Has this layer been working for your application?  I'm curious what might be the issue.

Best,
John

sandeepkumar-skb:
Hi @jaybdub, Thanks for inspecting. I got the same error result as well. However when I did a pSNR check of the plugin on the following test I got a pSNR : `nan`. `nan` being the best case scenario (no corruption of the signal)

```
import torch
import torch.nn as nn
import tensorrt as trt
import numpy as np

class GN(nn.Module):
    def __init__(self, num_groups, num_channels ):
        super().__init__()
        self.gn = torch.nn.GroupNorm(num_groups, num_channels)

    def forward(self, tensor):
        return self.gn(tensor)

num_channels = 128
num_groups = 8
inp = torch.randn((1, num_channels, 64, 2048), device='cuda')
my_gn = GN(num_groups, num_channels)
my_gn.cuda().eval()
gn_trt = torch2trt( my_gn, [inp], log_level=trt.Logger.INFO, fp16_mode=True)

output = my_gn(inp)
output_trt = gn_trt(inp)
output = output.cpu().detach().numpy().flatten()
output_trt = output_trt.cpu().detach().numpy().flatten()

MSE_ = np.sum(np.square(output - output_trt))/ len(output)
try:
    pSNR_ = 20 * math.log10(np.max(abs(output))) - 10 * math.log10(MSE_)
except:
    pSNR_ = "Nan"
print(f"MSE {MSE_} and pSNR {pSNR_}")
```

Thanks,
Sandeep

sandeepkumar-skb:
Hi @jaybdub ,
I also ran the tests in the converter separate in this script:
```
import torch
import torch.nn as nn
import tensorrt as trt
import numpy as np

class GN(nn.Module):
    def __init__(self, num_groups, num_channels, eps=1e-5):
        super().__init__()
        self.gn = torch.nn.GroupNorm(num_groups, num_channels, eps=eps)

    def forward(self, tensor):
        return self.gn(tensor)

num_channels = 10
num_groups = 2
inp = torch.randn((1, num_channels, 112, 112), device='cuda')
my_gn = GN(num_groups, num_channels, 1e-4)
my_gn.cuda().eval()
gn_trt = torch2trt( my_gn, [inp], log_level=trt.Logger.INFO, fp16_mode=True)

output = my_gn(inp)
output_trt = gn_trt(inp)
# compute max error
max_error = 0
max_error_i = 0
max_error_i = torch.max(torch.abs(output - output_trt))
if max_error_i > max_error:
    max_error = max_error_i
print(max_error)
```
I used the same metric that's used in https://github.com/NVIDIA-AI-IOT/torch2trt/blob/master/torch2trt/test.py#L42-L51
and I get an error of `0`.

Thanks!

jaybdub:
Hi @sandeepkumar-skb ,

Thanks for sharing these results.  Interesting, seems that the input shape was all that changed, I'm curious what the underlying issue is.

Anyways, let me test what you shared to see if I can reproduce on my machine.

Best,
John

jaybdub:
When I use the same data that was passed for conversion, I also get a max-error of 0.  However, when I instantiate new data.

```python
inp2 = 
output = my_gn(inp)
output_trt = gn_trt(inp)

**Comment by jaybdub** [comment]
Interestingly, it seems like subsequent calls to the TensorRT version match the output of the PyTorch module.

So just the first call to ``gn_trt(inp)`` produces different outputs.  For example, multiple calls to the same TensorRT module will produce different outputs.

```
inp = torch.randn((1, num_channels, 112, 112), device='cuda')
torch.max(torch.abs(gn_trt(inp) - gn_trt(inp)))
```

Will produce a non-zero value. Perhaps there is a synchronization issue.

**Comment by sandeepkumar-skb** [comment]
Hi @jaybdub ,
I think we have uncovered a pytorch bug in group_norm. Technically if the model is in eval mode `GroupNorm` should not be updating any statistics but seems like that's not the case unlike BatchNorm.
If you run the test like this:
```
    with torch.no_grad():
        outputs = module(*inputs)
        outputs_trt = module_trt(*inputs_trt)
```
then the `Max_error` is `0`.
Technically `torch.no_grad()` should just disable the autograd engine and `model.eval()` should be suffice for inference but it doesn't hurt to add `torch.no_grad()` as it will give memory savings. 
Either ways, `GroupNorm` should not be behaving differently with and without `torch.no_grad()` so I will create a PyTorch bug for this one.

For the current PR: 
I will update the PR use `with torch.no_grad()` in `test.py`

**Comment by jaybdub** [comment]
Hi @sandeepkumar-skb ,

I see in the PyTorch documentation that updating statistics for group norm is expected behavior in inference mode.  

> This layer uses statistics computed from input data in both training and evaluation modes.

From:

https://pytorch.org/docs/stable/generated/torch.nn.GroupNorm.html

Maybe this is expected behavior?


On another note, I'm actually seeing that the TensorRT version output changes for the same input.   Calling ``gn_trt(inp)`` will produce different results each call. Curious if you also see this.

Best,
John

**Comment by sandeepkumar-skb** [comment]
Oh, thanks for catching the `GroupNorm` info. Perhaps, its expected after all.

Re different results. its very weird. It seems very alternate run is producing correct result:
```
inp = torch.randn((1, num_channels, 224, 224), device='cuda')
with torch.no_grad():
    output = my_gn(inp)

for _ in range(100):
    with torch.no_grad():
        output_trt = gn_trt(inp)
    max_error = torch.max(torch.abs(output - output_trt))
    print("max_error: ", max_error)
```
Results:
```
max_error:  tensor(0., device='cuda:0')
max_error:  tensor(4.6691, device='cuda:0')
max_error:  tensor(0., device='cuda:0')
max_error:  tensor(4.6691, device='cuda:0')
max_error:  tensor(0., device='cuda:0')
max_error:  tensor(4.6691, device='cuda:0')
max_error:  tensor(0., device='cuda:0')
max_error:  tensor(4.6691, device='cuda:0')
max_error:  tensor(0., device='cuda:0')
max_error:  tensor(4.6691, device='cuda:0')
max_error:  tensor(0., device='cuda:0')
```

**Comment by jaybdub** [comment]
I think I found the underlying bug.

After calling ``at::group_norm`` , the result is never copied to the TensorRT output buffer.  So the values of the output buffer are whatever was there when allocated.

Adding the following works for me.

```
at::Tensor output_tmp = at::group_norm(input, num_groups, weight_t, bias_t, eps=eps);
output.copy_(output_tmp);
```

This sets the output value.  I now see that the TensorRT module produces results identical to the PyTorch.

Unfortunately, I can't find an in-place version of this function with at, so there is an extra copy needed.  This may have a performance hit, but at least the model will be enabled.

I've also made some changes in my working branch that simplify the serialization / deserialization a bit (Using at::Tensor directly instead of vector<double>).  

Best,
John

**Comment by sandeepkumar-skb** [comment]
Thanks @jaybdub for catching that. My bad - I thought it happening in-place.

Although after making that change I get consistent results but `max_error` is still high.
Curious if you are seeing the same?

Thanks,
Sandeep

**Comment by jaybdub** [comment]
I have some other changes, with those, I see 0 error

```python
| torch2trt.converters.group_norm.test_group_norm_trt_g2_fp32 | float32 | [(1, 10, 112, 112)] | {} | 0.00E+00 | 7.55e+03 | 3.47e+03 | 0.177 | 0.254 |
| torch2trt.converters.group_norm.test_group_norm_trt_g2_eps_fp32 | float32 | [(1, 10, 112, 112)] | {} | 0.00E+00 | 7.57e+03 | 5.62e+03 | 0.177 | 0.256 |
```

I can open a new PR with these changes applied to your branch if that works for you to test.

Best,
John

sandeepkumar-skb:
Absolutely!

jaybdub:
@sandeepkumar-skb 

The changes are pushed here https://github.com/NVIDIA-AI-IOT/torch2trt/pull/437.

Please give it a try and let me know how it goes, I think we can continue discussion on that PR.

Best,
John

sandeepkumar-skb:
Closing this PR as the plugin is merged! See : [#437 ](https://github.com/NVIDIA-AI-IOT/torch2trt/pull/437)

PR #442, opened 2020-11-07
Converter for Nvidia quantization aware training library

SrivastavaKshitij:
## QAT support using Nvidia quantization library 

This PR provides converter for QAT layers provided by  nvidia quantization aware training library present [here](https://github.com/NVIDIA/TensorRT/tree/master/tools/pytorch-quantization)

Please refer to `README.md` under `torch2trt/examples/` for complete workflow. [Link](https://github.com/NVIDIA-AI-IOT/torch2trt/blob/2926b727daee9b56ace18060d67b40a18b5029b1/examples/quantization_aware_training/README.md)

## Workflow Explanation 

1. Nvidia QAT library quantizes the input to various operators. 
2. TensorRT 7.xx release notes provides very basic support for QAT and the requirements are that the dynamic range should be provided for all the layers that are to be executed at INT8. [Link](https://docs.nvidia.com/deeplearning/tensorrt/developer-guide/index.html#enable_int8_python)

3. Given TensorRT requirement and torch2trt workflow that supports pytorch eager mode, we have to quantize the output of operators as opposed to input to the operator. 

For e.g.

NVidia Qat Library does the following 

```
Input --> Activation Quantization --> Conv (weight quantization) --> BN --> ReLU 
```

In the above workflow, since no dynamic range is provided for ReLU , it will not convert to INT8. 

Therefore , I rewrote layers so that the output of layers are quantized instead of input. 

```
Input --> Conv (weight quantization) --> BN --> ReLU (activation quantization) -->
```

However, if we look at the converter for BN in torch2trt , it maps to a `TRT scale layer`. Since we don't provide a dynamic range for `scale layer`, therefore , TRT throws the entire combination of  `conv-->bn to `FP16`. 

Hence, layers that I created under `torch2trt/qat_layers/quant_conv.py` folds BN into Conv and then performs quantization. So, if you look at the converter, `converters/QuantConvBN.py`, it doesn't require a scale layer and mapping satisfies TRT conditions. 

4. Custom converters are working, therefore, original converters for `conv,bn,relu` have not been touched. 

5. Every layer under `qat_layers/quant_{}.py` has a corresponding version for inference. This is required because Nvidia QAT library performs quantization as a part of its `forward` pass , that means, all the operations required for quantizing the inputs / weights will end up in the TRT engine which we dont want. We only want the training to provide us with dynamic range so that we can map the range to trt layers when generating an engine.

## TensorRT 7.xx limitations for QAT

1. TRT only supports per tensor quantization (per channel quantization is not currently supported)
2. TRT only supports summetric quantization with zero point always 0. That means activation quantization can only be between [0,127] 

## Results

I have compared results for FP32, FP16, INT8 (PTC and QAT) under `torch2trt/examples/quantization_aware_training/README.md`

Results for Resnet 18 are as expected. However, Resnet34 QAT performs worse than PTC. 

It could be due to the following reasons. 

1. Not every layer is robust to INT8 quantization. Maybe, I am quantizing a layer in Resnet34 that calibration leaves out , causing a drop in accuracy. Also, we dont know which layers are converted to INT8 during calibration as it is a black box. 

2. Lack of per-channel quantization support for weights in QAT could also be a possibility, which apparently, is not a problem in calibration. **I am doing some more experiments to figure out the root cause**

## Next Step

In this PR , I have converted widely used operators -  `Conv, BN and ReLU` to INT8 . 

Next PR will introduce more operators such as various types of pooling and linear layer.


jaybdub:
Hi @SrivastavaKshitij ,

Thanks for this PR.  Looks very interesting!

It may take me some time to review.  Here are some immediate thoughts in case I forget or if you have any comments.  Definitely not requested actions.  

1.  Would be cool to have a documentation table detailing accuracy of FP32, FP16, INT8 (calibration), INT8 (QAT) for the toy dataset.   
2.  I notice you use the existing Conv converter for the Quant layers.  
    1.  How would this behave for a mixed model (some Quant layers, some regular Convs)?  Is there any point in supporting such a configuration / could this arise in practice?
    2.  Would the following PR (or similar change) which allows you to create converters for non-torch layers have helped here?  https://github.com/NVIDIA-AI-IOT/torch2trt/pull/175

Let me know if you have any thoughts on the above.

Looking forward to trying this out.

Best,
John

SrivastavaKshitij:
Hey @jaybdub 👍 

1. I was planning to include and overhaul the README.md page for the library but I didnt do it right away is because I wanted to get the PR out so that you have time to look at it. I was planning to work with you to tag a release of Torch2trt version 2.0 with QAT support and also overhaul README.md page. I think this library has grown a lot since the previous release and we should show that on the main readme page.

2. Yes, you are correct. I implemented the mixed model for training but somehow it just skipped my mind for the converter part. Your suggestion looks good. I am in the middle of bringing more changes to this PR and will incorporate your suggestion to create a converter for custom layer. I didnt know that could be done before, so thanks for pointing it out. 

Stay tuned to more updates on this PR

jaybdub:
Awesome!  Yes thanks for sharing, it's nice to get the early look.  Look forward to updates.

I think you're right, it's about time for a version increment / release.   This would be a great example to include / document.

FYI, here are some thoughts I have on integration / release in case it helps while you're making changes

1.  Rename 'notebooks' -> 'examples'.  
    - So this will read examples/image_classification, examples/image_segmentation.
2.  Move ``torch2trt.examples`` -> ``examples/quantization_aware_training``
3.  Define a custom converter for Quant layers inside the example using (2) or a similar method.  
    >  I'm happy to work on enabling support for non-torch custom converters for this.  This feature has been requested before for other purposes, but was delayed because I was concerned with the maintenance liability.  QAT use case would be a great test of the feature.
4. Remove ``qat_mode`` from ``torch2trt(...)``.  I think it should no longer be needed if customer converter works.  Please correct me if I'm wrong on this. 
5.  For documentation.  We could reference the examples/quantization_aware_training/README from the main README.  We could also potentially add a documentation page examples/quantization_aware_training.  https://nvidia-ai-iot.github.io/torch2trt/master/

Let me know what you think, or if this makes sense to you.  

Best,
John

SrivastavaKshitij:
All good ideas John. I am incorporating all these changes in my updates.

SrivastavaKshitij:
@jaybdub : PR that you mentioned about using custom converter is still open , it was not merged into master

SrivastavaKshitij:
I am having some problem integrating @tensorrt_method

jaybdub:
Hi @SrivastavaKshitij ,

The mechanism behind ``tensorrt_method`` is actually a bit different from ``tensorrt_converter``.  It requires additional changes to the conversion flow than you currently have incorporated.  You can see these here:

https://github.com/NVIDIA-AI-IOT/torch2trt/pull/175/files

Please note, even with these changes, tensorrt_method is still experimental.  I still need to do more tests, and consider how this would impact torch2trt users.  There is a slight chance it may prove problematic.  I'll try to rebase the other PR against master and pick this up.

For now, you may be able to try

```bash
@tensorrt_converter('torch2trt.qat_layers.quant_linear.IQuantLinear.forward`)
def convert_quant_linear(ctx):
    # ...
```

This may conveniently work if ``torch2trt.qat_layers.quant_linear`` is imported inside ``torch2trt/torch2trt.py``

```python
import torch2trt.qat_layers.quant_linear
```

In order for ``tensorrt_converter`` to wrap a function, torch2trt must have the module imported.  This is because the ``tensorrt_converter`` mechanism cannot wrap the function if it's unaware it exists.  This approach isn't as ideal, but may enable you to test the QAT converter.

Please let me know if this helps or you have any questions / concerns.  I'll pick up the other PR and let you know of any updates.

Best,
John

SrivastavaKshitij:
Ok John. I will use the hack for now and complete testing my implementation.

SrivastavaKshitij:
Hey @jaybdub : Custom converter doesnt work. Whenever Torch2trt sees a conv layer , it point to `Conv.py` instead of `QuantConv.py` and then it fails to convert to INT8 because `Conv.py` doesnt have dynamic range. Could you please help me in figuring out why the custom converter is not being picked up during conversion 

Thanks

jaybdub:
Sure, cloning your PR to take a look now.

SrivastavaKshitij:
Thank you. If you have any difficulty, please let me know.

jaybdub:
Hi @SrivastavaKshitij ,

As a debugging step, I tried dummy converters using ``pytorch_quantization`` layers directly.

```python
from torch2trt import *
import torch
import torch.nn as nn
import torch.nn.functional as F

from pytorch_quantization.nn import QuantConv2d, QuantLinear
from pytorch_quantization.tensor_quant import QuantDescriptor

@tensorrt_converter('torch.nn.Conv2d.forward')
def convert_Conv2d(ctx):
    print('In Conv2d')
    
@tensorrt_converter('torch.nn.Linear.forward')
def convert_Linear(ctx):
    print('In Linear')

@tensorrt_converter('pytorch_quantization.nn.QuantConv2d.forward')
def convert_QuantConv2d(ctx):
    print('In QuantConv2d')
    
@tensorrt_converter('pytorch_quantization.nn.QuantLinear.forward')
def convert_QuantLinear(ctx):
    print('In QuantLinear')
    
    
linear = nn.Linear(50, 10).cuda().eval()
conv = nn.Conv2d(1, 10, kernel_size=5).cuda().eval()
quant_conv = QuantConv2d(1, 10, kernel_size=5).cuda().eval()
quant_linear = QuantLinear(50, 10).cuda().eval()

linear_data = torch.zeros(1, 50).cuda()
conv_data = torch.zeros(1, 1, 28, 28).cuda()

try:
    _ = torch2trt(conv, [conv_data])  # prints "In Conv2d"
except:
    pass
try:
    _ = torch2trt(linear, [linear_data ])  # prints "In Linear"
except:
    pass
try:
    _ = torch2trt(quant_conv, [conv_data])  # prints "In QuantConv2d"
except:
    pass
try:
    _ = torch2trt(quant_linear, [linear_data ])  # prints "In QuantLinear"
except:
    pass
```

These seem to print the correct associated converter.  Please note, they throw errors after printing, since they are dummy converters.

I haven't yet figured what is happening in your instance.

I'm curious, are ``torch2trt.qat_layers`` different from ``pytorch_quantization`` Quant layers?  Or could we possibly attach the converters directly to the ``pytorch_quantization`` layers as done above?  

Right now it works because you have manually imported ``pytorch_quantization`` in ``torch2trt.py``.  But we could potentially add custom imports to ``tensorrt_converter`` to support custom libraries like this

```python
@tensorrt_converter('pytorch_quantization.nn.QuantConv2d.forward', imports=['pytorch_quantization'])
def convert_QuantConv2d(ctx):
    # ...
```

Let me know if you have any thoughts on the above.  I'll keep investigating your branch in the meantime.

Best,
John

SrivastavaKshitij:
Hey John, 

Thanks for looking into it. 

1. Could there be a possibility that because `torch2trt.qat_layers.quant_conv` is present inside torch2trt folder , it's not registering probably. I highly doubt it but just wanted to get your opinion

2. The most common combination that is being used for qat is `Conv-> BN-> ReLU` . 

Nvidia quantization library quantizes the input of conv instead of output of ReLU. 

So , when nvidia quantizes that combination it looks like this

`Quantized I/P  ----> Conv (Quantized weights) ----> BN ------> ReLU` 

However, TRT complains that it cannot convert ReLU in the combination to INT8 because the dynamic range is missing for the layer. 

So , I removed input quantizer from Conv and put it after ReLu and those changes are reflected under `torch2trt/qat_layers/quant_{}.py`

jaybdub:
Regarding (1).  I noticed the IQuant layers do not override the ``forward`` method.  I believe this nuance is causing the issue using ``tensorrt_converter``

```python
class NoOverrideConv2d(nn.Conv2d):
    pass

class OverrideConv2d(nn.Conv2d):
    def forward(self, *args, **kwargs):
        return super(OverrideConv2d, self).forward(*args, **kwargs)

print(NoOverrideConv2d.forward)
print(OverrideConv2d.forward)
```    

outputs

```
<function Conv2d.forward at 0x7f979d8c1b70>
<function OverrideConv2d.forward at 0x7f977ad27bf8>
```

As you can see, for the class that does not override the ``forward`` method, it returns the original Conv2d method.

Perhaps this is the issue, but I'm not 100% sure at the moment.

Best,
John

SrivastavaKshitij:
Ahh man. How did I miss that. Thanks for pointing that out . I will make changes.

SrivastavaKshitij:
@jaybdub : Actually that doesnt work and here is the reason .

Nvidia quantization library performs quantization of weights and activations in the forward pass. If were to use that library then all the functions that are used in qat library will become part of the model which is not ideal. 

That's why I created IQuantConv2d, IQuantReLU etc. The reason behind these classes for inference is that I will create dummy tensor for weight and activation quantized tensors so that the original values can be loaded from the ckpt and the forward pass will be the original conv forward pass. As I am carrying the quantized values with me , I will just map it to the TRT layer during conversion. 

So there are two ways to do it. 

1. Use the Conv converter and add the following code there:

```
if module.groups is not None:
        layer.num_groups = module.groups

    if hasattr(module,'_weight_quantizer'):
        w_quant_amax = module._weight_quantizer.learned_amax
        print(w_quant_amax)
        layer.precision = trt.int8
        layer.set_output_type(0,trt.int8)
        conv_out = layer.get_output(0)
        conv_out.dynamic_range=(-w_quant_amax,w_quant_amax)

    output._trt = layer.get_output(0)

```

`hasattr` will work for both the original conv2d layers and conv2d layers + quantized weights. 

2. I can still use the `QuantConv.py` converter and I can call `Conv.py` converter inside it and then assign dynamic range.

Pseudo code:

Inside `QuantConv.py`  : 

```
@tensorrt_converter('torch2trt.qat_layers.quant_conv.IQuantConv2d.forward')
def convert_QuantConv(ctx):
       convert_Conv_trt7(ctx)
       `take the output from the above converter and assign dynamic range`
       w_quant_amax = module._weight_quantizer.learned_amax
        print(w_quant_amax)
        layer.precision = trt.int8
        layer.set_output_type(0,trt.int8)
        conv_out = layer.get_output(0)
        conv_out.dynamic_range=(-w_quant_amax,w_quant_amax)

       
```

jaybdub:
Hi @SrivastavaKshitij ,

Thanks for the explanation / proposals!

From what I currently understand, method #2 looks best to me.  Using the ``convert_Conv_trt7`` converter, or even copying the contents of the conv converter into Quant converter seems like a good approach. 

I like this because it isolates the functionality of this quantization approach from other core components.  This provides some flexibility in how it's distributed and potentially modified / extended.   For example, if you later find a different quantization approach you like, you can readily apply that without worrying about needing to conform to a previous convention or modify a built-in torch2trt converter.  This also allows you to define the QAT layers / converters directly inside the example itself if needed.

Let me know if you have any thoughts.

Best,
John

SrivastavaKshitij:
@jaybdub : If you look at the latest commit , I made all the changes . However, it still calls `Conv.py` converter instead of `QuantConv.py`

I was able to recreate this issue with your example

```
from torch2trt.torch2trt import *
import torch
import torch.nn as nn
import torch.nn.functional as F

from pytorch_quantization.nn import QuantConv2d, QuantLinear
from pytorch_quantization.tensor_quant import QuantDescriptor
from torch2trt.qat_layers.quant_conv import IQuantConv2d

print(IQuantConv2d.forward)

@tensorrt_converter('torch.nn.Conv2d.forward')
def convert_Conv2d(ctx):
    print('In Conv2d')

@tensorrt_converter('torch.nn.Linear.forward')
def convert_Linear(ctx):
    print('In Linear')

@tensorrt_converter('pytorch_quantization.nn.QuantConv2d.forward')
def convert_QuantConv2d(ctx):
    print('In QuantConv2d')

@tensorrt_converter('pytorch_quantization.nn.QuantLinear.forward')
def convert_QuantLinear(ctx):
    print('In QuantLinear')

@tensorrt_converter('torch2trt.qat_layers.quant_conv.IQuantConv2d.forward')
def convert_quantconv2d(ctx):
    print("In IQuantConv2d")


linear = nn.Linear(50, 10).cuda().eval()
conv = nn.Conv2d(1, 10, kernel_size=5).cuda().eval()
quant_conv = QuantConv2d(1, 10, kernel_size=5).cuda().eval()
quant_linear = QuantLinear(50, 10).cuda().eval()
iquant_conv = IQuantConv2d(1,10,kernel_size=5).cuda().eval()

linear_data = torch.zeros(1, 50).cuda()
conv_data = torch.zeros(1, 1, 28, 28).cuda()


try:
    _ = torch2trt(conv, [conv_data])  # prints "In Conv2d"
except:
    pass
try:
    _ = torch2trt(linear, [linear_data ])  # prints "In Linear"
except:
    pass
try:
    _ = torch2trt(quant_conv, [conv_data])  # prints "In QuantConv2d"
except:
    pass
try:
    _ = torch2trt(quant_linear, [linear_data ])  # prints "In QuantLinear"
except:
    pass

try:
    _ = torch2trt(iquant_conv, [conv_data]) # prints "in IQuantConv2d")
except:
    pass

```

Output

```
<function IQuantConv2d.forward at 0x7f9b20b6e510>
In Conv2d
In Linear
In QuantConv2d
In QuantLinear
In Conv2d
```

jaybdub:
@SrivastavaKshitij FYI,

Happy new year!

I've added support for custom keyword arguments to torch2trt in this PR

https://github.com/NVIDIA-AI-IOT/torch2trt/pull/482

Please let me know if this works for your purposes.

Best,
John

SrivastavaKshitij:
Hey John, I will look into it. I am also planning to (next week or so) to make changes to this PR so that my QAT stuff is under `torch2trt/contrib` folder and then we can go ahead and merge it.

PR #488, opened 2021-01-13
wheel build fails in docker image due to nvidia runtime not available

fracpete:
If the library gets cloned/installed inside a docker image, the nvidia runtime isn't available, resulting in the following error:
```
OSError: libcurand.so.10: cannot open shared object file: No such file or directory
```
This stems from the `torch` import in the `setup.py` file, which will attempt to load the CUDA libraries.

By making the torch import/use conditional to the `--plugins` flag, it is possible to install the library inside a docker image (though without the plugins, of course).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,5 @@
 import sys
-import torch
 from setuptools import setup, find_packages
-from torch.utils.cpp_extension import BuildExtension, CUDAExtension, CppExtension
 
 def trt_inc_dir():
     return "/usr/include/aarch64-linux-gnu"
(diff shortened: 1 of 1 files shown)


jaybdub:
Hi @fracpete,

Thanks for reaching out!

I believe setting the default docker runtime to "nvidia" will resolve this problem.  

Simply edit ``/etc/docker/daemon.json`` and add ``"default-runtime": "nvidia",``

You can find more details here https://docs.nvidia.com/dgx/nvidia-container-runtime-upgrade/index.html#using-nv-container-runtime.

I believe this may require restarting the docker daemon afterwards.  ``sudo systemctl restart docker``

Please let me know if this helps or you run into issues.

Best,
John

fracpete:
Setting the default runtime environment unfortunately requires you to have an actual NVIDIA card on the machine that builds the image, which is not the case for our build server (or my laptop).

PR #529, opened 2021-03-20
Update dummy_converters.py

JonathanSKent:
Just changes "is not" to "!=", which otherwise throws a warning.


jaybdub:
This is now merged.  Thanks @JonathanSKent !

PR #585, opened 2021-07-06
[DO NOT MERGE][Reproducing issue] : View converter , bs >1

SrivastavaKshitij:
Hi @jaybdub

I recently came across a use case where the conversion fails for view converter when batch size >1 . 

Command: `python -m torch2trt.test --name=view` 

Some pointers:

1. I have disabled all other tests except one to reproduce the error that I am seeing.
2. One interesting thing that I noticed was that even though I was using an input tensor size of `[2,3,3,3]` , view converter was still seeing the input size as `[1,3,3,3]` . Line 15 in the `view` converter. I couldnt de-bug it though.


jaybdub:
Hi @SrivastavaKshitij ,

Thanks for pointing this out.

This likely has to do with this line

https://github.com/NVIDIA-AI-IOT/torch2trt/blob/817f937be44627f81d3efed9fafd4f8a905e5916/torch2trt/torch2trt.py#L517

When we pass the example data into the model, we do it with batch size 1.

From my understanding, converters shouldn't modify the batch dimension, so this would be acceptable, but perhaps I'm missing something.

Let me know if this helps.

Best,
John

SrivastavaKshitij:
Hey John, 

Did some more experiments. 

when I changed line 517 from 

`inputs = [tensor.clone()[0:1] for tensor in inputs]`

to 

`inputs = [tensor.clone() for tensor in inputs]`

I was able to see the batch dimension. So that's good. However, 

https://github.com/NVIDIA-AI-IOT/torch2trt/blob/817f937be44627f81d3efed9fafd4f8a905e5916/torch2trt/torch2trt.py#L153

removes the batch dimension again and I dont know why is that. What do you think ? 

I added a print statement of the tensor shape before and after line 157 and this is what I got 

```
torch.Size([2, 3, 3, 3]) 
(3, 3, 3)
```

Somehow, batch dim is removed. 

Now this may not have impacted other ops but it will impact some of the ops such as view because the volume of the tensor will not match 

```
[TensorRT] ERROR: [SHUFFLE #1] torch.Tensor.view(tensor(shape=[2, 3, 3, 3], dtype=torch.float32), 1, -1): volume mismatch. Input dimensions [3,3,3] have volume 27 and output dimensions [54] have volume 54.
```

torch2trt/converters/view.py
```diff
@@ -45,13 +49,13 @@ def forward(self, x):
...
 
 
-@add_module_test(torch.float32, torch.device('cuda'), [(1, 3)])
-@add_module_test(torch.float32, torch.device('cuda'), [(1, 3, 3)])
-@add_module_test(torch.float32, torch.device('cuda'), [(1, 3, 3, 3)])
+#@add_module_test(torch.float32, torch.device('cuda'), [(1, 3)])
+#@add_module_test(torch.float32, torch.device('cuda'), [(1, 3, 3)])
+@add_module_test(torch.float32, torch.device('cuda'), [(2, 3, 3, 3)],max_batch_size=3)
 def test_view_1d():
     return View(1, -1)
```
jaybdub, line 56: It looks like the test case is hard-fixing the batch dimension to 1 here.  

In general, TensorRT engines are broadcast across the batch dimension, so operations that change the batch dimension aren't permitted.

Perhaps adding a test case with View(2, ...) would work for batch size to.  Or maybe even View(-1, ...) with other dimensions specified in full.

jaybdub, line 56: It looks like the test case is hard-fixing the batch dimension to 1 here.  

In general, TensorRT engines are broadcast across the batch dimension, so operations that change the batch dimension aren't permitted.

Perhaps adding a test case with View(2, ...) would work for batch size to.  Or maybe even View(-1, ...) with other dimensions specified in full.

PR #599, opened 2021-07-27
[WIP] Trt 8

jaybdub:
# Changes

- uses builder config class instead of directly setting builder attributes
- updates interpolate layer to use new API for align_corners

# Known issues

- boolean operators (>, <, ...) are no longer working and generate the following error

    ```bash
    Internal Error (Assertion kCudaElementWiseSupportsBool failed.Cuda backend does not support kEQUAL/kGREATER/kLESS)
    ```

### DIFF
--- a/torch2trt/converters/interpolate.py
+++ b/torch2trt/converters/interpolate.py
@@ -86,7 +86,11 @@ def convert_interpolate_trt7(ctx):
         layer.resize_mode=trt.ResizeMode.NEAREST
 
     if align_corners != None:
-        layer.align_corners = align_corners
+        if trt_version() > '8.0':
+            if align_corners:
+                layer.coordinate_transformation = trt.ResizeCoordinateTransformation.ALIGN_CORNERS
+        else:
+            layer.align_corners = align_corners
 
     output._trt = layer.get_output(0)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Master]
 
+- Added support for TensorRT 8
+
 ## [0.3.0] - 07/15/2021
 
 - Added converter for ``torch.nn.functional.adaptive_avg_pool3d``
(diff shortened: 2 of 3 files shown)


torch2trt/converters/interpolate.py
```diff
@@ -86,7 +86,11 @@ def convert_interpolate_trt7(ctx):
         layer.resize_mode=trt.ResizeMode.NEAREST
 
     if align_corners != None:
-        layer.align_corners = align_corners
+        if trt_version() > '8.0':
+            if align_corners:
```
JWLee89, line 90: Just out of curiosity, since we already perform the following check: 

```
if align_corners != None
```
Wouldn't the following `if align_corners` be redundant?

jaybdub: Hi @JWLee89,  thanks for pointing this out. I'm not sure if it happens in practice, but I think if the user passes ``False`` the second check is needed.  There might be a cleaner way to organize this logic.  Let me know if you see any issues.

JWLee89: Hi @jaybdub ,

Thank you for the response. 

If the goal is to filter out `None` and `False`,  I think we can replace the following code block 

```python
if align_corners != None:
    layer.align_corners = align_corners
    if trt_version() > '8.0':
        if align_corners:
```

With the following 

```Python
# replace if align_corners != None to filter out both None and False
if align_corners:
    layer.align_corners = align_corners
    if trt_version() > '8.0':
        # enter remaining logic ...
```

The reason is because `if align_corners` automatically filters out values that are `None` and `False`. I have added a simple code snippet below to prove that. 

```Python
    for align_corners in [False, None, True]:
        # align corners filters out both False and None
        if align_corners:
            # Truthy only called on "True"
            print(f"Truthy! Passed in: {align_corners}")
        else:
            print(f"Falsy! Passed in: {align_corners}")
```

Of course, I might have missed out on some internal implementation detail-specific use case or taken things out of context. If so, please feel free to disregard this, as removing / adding a for loop is not too critical. 

Thank you for taking time to read through this!

PR #607, opened 2021-08-11
Added TensorRT Converters for Basic Casting Operations

JWLee89:
Added support for the following casting operations under the `converters/cast.py`.
- `.float()`. e.g `x.float()`
- `.int()`. e.g `x.int()`
- `.bool()`. e.g `x.bool()`

Tests are included along with the converters (see attached image below).

**Note**: `.long()` is not included as this is a 64 bit data type and TensorRT does not support 64 bit data types. 

cast_test

### DIFF
--- a/torch2trt/converters/cast.py
+++ b/torch2trt/converters/cast.py
@@ -32,6 +32,8 @@ def convert_bool(ctx):
 def convert_bool(ctx):
     convert_cast(ctx)
 
+# Used for torch.Tensor.<cast> tests
+# --------------------------------------------
 
 class TorchFloat(torch.nn.Module):
     def __init__(self):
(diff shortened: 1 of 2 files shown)


torch2trt/converters/cast.py
```diff
@@ -0,0 +1,72 @@
...
+def convert_cast(ctx):
+    """
+    A simple converter for supporting casting operations.
+
+    IMPORTANT: Note that because TensorRT does not support
+    64 bit data types, .long() will not be supported
+    """
+    input_tensor = ctx.method_args[0]
+    layer = ctx.network.add_identity(input_tensor._trt)
+    output = ctx.method_return
```
SrivastavaKshitij, line 14: In all the cases, IMO, all the layers are getting converted to an identity layer at FP32. 

This might not work for lower precisions such as INT8 and FP16

We need something like this 

`layer.precision = trt.int8` for int8 and something similar for fo16

JWLee89: Thank you for the useful feedback.  

You are right. I did not factor in lower precisions (INT8 and FP16). 
I will update the code, run the tests and report back after committing.

JWLee89: @SrivastavaKshitij I have made the suggested updates. If you have any additional suggestions / recommendations, I will take those into consideration and make the appropriate changes. 

Thank you for taking time to make a review.

JWLee89:
Hi @jaybdub, 

I was wondering if there are any possible potential issues with this PR? 
If so, do you have any suggestions on how to improve the feature?

P.S. Apologies for the ping. I know that you are busy. 

Thank you very much for taking time out of your schedule to take a look at this. 

Best regards,
Jay

PR #629, opened 2021-09-15
Add support for inplace mul_ operator from pytorch

andi4191:
Hi,

This change allows us to parse and convert torch.Tensor.mul_ operator (in place version of mul)
https://pytorch.org/docs/stable/generated/torch.Tensor.mul_.html

Signed-off-by: Anurag Dixit <a.dixit91@gmail.com>


jaybdub:
Thanks @andi4191 !

PR #633, opened 2021-09-21
Add torch.clone to converters.

chaoz-dev:
Adds `torch.clone` to the list of available converters. 

Test with:
```
python3 -m torch2trt.test --name clone
```


chaoz-dev:
I've also included some failed test cases; let me know if there are other changes needed!

PR #638, opened 2021-09-28
Fix stack converter to support negative dims.

chaoz-dev:
torch.stack can take negative dims, which need to be reversed when input into TensorRT, which cannot support negative dims. (This step is automatically performed if using ONNX).


torch2trt/converters/stack.py
```diff
@@ -23,6 +27,7 @@ def convert_cat_trt7(ctx):
     layer.axis = dim - 1
```
chaoz-dev, line 27: Just to confirm -- this will be invalid if converting `torch.cat(..., 0)` (ie. if we try to cat in the first dimension); however, because we expect that the first dimension is always the batch dimension, we accept this limitation, correct? (Note that this PR with the negative dims preserves this behavior).

PR #657, opened 2021-11-10
Use common helper function set_layer_precision instead of private one for converters.

chaoz-dev:
Several converters were using `_set_layer_precision`, which was a stand-in while a common one could be landed. Since the common one landed, we can now use that function instead of having to copy a private version for each converter. The common `set_layer_precision` can be found in `torch2trt.py`.

### DIFF
--- a/torch2trt/converters/clone.py
+++ b/torch2trt/converters/clone.py
@@ -2,27 +2,6 @@ from torch2trt.torch2trt import *
 from torch2trt.module_test import add_module_test
 
 
-def _set_layer_precision(ctx, layer):
-    # Supported TRT precisions as given by torch2trt_kwargs.
-    INT8_MODE = "int8_mode"
-    FP16_MODE = "fp16_mode"
-
-    # Check that args exist as expected in torch2trt_kwargs.
-    trt_kwargs = ctx.torch2trt_kwargs
-    assert INT8_MODE in trt_kwargs
-    assert FP16_MODE in trt_kwargs
-
-    is_int8 = trt_kwargs.get(INT8_MODE, False)
-    is_fp16 = trt_kwargs.get(FP16_MODE, False)
-
-    if is_int8:
-        layer.precision = trt.int8
-        layer.set_output_type(0, trt.int8)
-    elif is_fp16:
-        layer.precision = trt.float16
-        layer.set_output_type(0, trt.float16)
-
-
 @tensorrt_converter('torch.clone')
 @tensorrt_converter('torch.Tensor.clone')
 def convert_clone(ctx):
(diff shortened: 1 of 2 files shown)


jaybdub:
Thanks @chaoz-dev !

Jae-1409:
Hi @chaoz-dev, 

Can I run partial quantization via 'set_layer_precision'?

I tried to set some layers' precision with the code below.

The engine was compiled well, but when I tested the inference speed on Xavier AGX board, it was even slower than the fp16 precision model.

    if partial:
        config.set_flag(trt.BuilderFlag.FP16)
        aaa = '1 14 15 26 27 36 37 44 45 52 53 60 61 68 69 76 77 86 87 98 99 112 113 114 115 116 117 118 119'.split( )
        count = 0
        for i in range(network.num_layers):
                if 'CONCATENATION' not in network.get_layer(i).name:
                    if f'CONVOLUTION #{aaa[count]}' in network.get_layer(i).name:
                        set_layer_precision(network.get_layer(i), 'fp16')
                        count += 1
                        if len(aaa)==count:
                            break

    if not 'qat_mode' in kwargs:
            calibrator = DatasetCalibrator(inputs, int8_calib_dataset, batch_size=int8_calib_batch_size, algorithm=int8_calib_algorithm)
            config.int8_calibrator = calibrator

Am I doing right?

And is there any **python** API to run partial quantization?

Layer-wise precision setting APIs seems to be used in C language in this [repo](https://github.com/NVIDIA/TensorRT/tree/main/samples/sampleINT8API)..

chaoz-dev:
AFAIK we don't currently support partial quantization in torch2trt at the moment; it should be all or nothing, depending on the args that are passed in. 

With respect to setting layer precision, at what point in the original code are you resetting the layer precision? 
Also note from the TRT documentation:
> Setting the precision constraint hints to TensorRT that it should select a layer implementation whose inputs and outputs match the preferred types, inserting reformat operations if the outputs of the previous layer and the inputs to the next layer do not match the requested types. Note that TensorRT will only be able to select an implementation with these types if they are also enabled via the flags in the builder configuration.

Overall though, it's a bit hard to tell exactly where the bottleneck is here just from the code (I think); you might need to profile so that we can find the issue.

Jae-1409:
@chaoz-dev , thank you for your response. 

I added my code at torch2trt/torch2trt/torch2trt.py line 668.

As I couldn't get access to each layer's information after trt-engine was built, I tried to change the network's (constructed by ConversionContext & ConversionHook) precision flag before engine compilation.

I changed only convolution layers' precisions (int8->fp16) and this maybe made the bottleneck as trt fails to fuse 'conv-bn-relu'.

PR #627, opened 2021-09-11
[WIP] Update plugin classes to match the base IPluginV2 class in TensorRT8

chitoku:
Updated the plugin classes to match the latest IPluginV2 base class in TnesorRT8, which is shipped with JetPack 4.6.
https://docs.nvidia.com/deeplearning/tensorrt/api/c_api/classnvinfer1_1_1_i_plugin_v2.html#af502120d140358f8d8139ade5005b9f5

Please note that this is not yet verified.

PR #660, opened 2021-11-10
Use kLinear instead of kNCHW for PluginFormat.

chaoz-dev:
Starting from TRT 8.x, `PluginFormat` deprecates `kNCHW` for `kLINEAR`. Update all usages to support new enum.


chaoz-dev:
This is based off of the original changes from https://github.com/NVIDIA-AI-IOT/torch2trt/pull/627

PR #691, opened 2022-02-04
Explicit batch and Einsum converter

jaybdub:


torch2trt/converters/cat.py
```diff
@@ -16,7 +16,7 @@ def convert_cat(ctx):
     trt_inputs = broadcast_trt_tensors(ctx.network, trt_inputs, len(output.shape) - 1)
```
chaoz-dev, line 16: Just pattern matching, but do we need to remove the `- 1` here? ie. 
```
    trt_inputs = broadcast_trt_tensors(ctx.network, trt_inputs, len(output.shape))
```
I assume the dim reduction here was also due to the previous removal of the batch dim.

torch2trt/torch2trt.py
```diff
@@ -534,6 +537,18 @@ def torch2trt(module,
...
+    
+    for i, name in enumerate(input_names):
+        # use fixed optimization profile
+        profile.set_shape(
+            name,
+            tuple(inputs[i].shape),
+            tuple(inputs[i].shape),
+            tuple(inputs[i].shape)
+        )
+        config.add_optimization_profile(profile)
```
chaoz-dev, line 550: Looks like this is for dynamic shapes? Is this necessary at the moment? I believe the current API will only support one (static) shape anyway, so I think we don't need _any_ optimization profiles (ie. we'll end up just using the given shape as before).

SrivastavaKshitij:
Hey @jaybdub : I checked this PR with different Pytorch NGC Containers and it is working fine for standard models such as mobilenets, resnets, efficientnets with different versions of pytorch and tensorrt. 

In my case, I didnt have dynamic shapes, however, we need to resolve the issue @chaoz-dev mentioned above. 

I think explicit batch is a great feature and should be landed soon 

thanks

jaybdub:
Thanks for the feedback on this PR @chaoz-dev  and @SrivastavaKshitij .

I've tested that the modification still pass all test cases.  Please let me know if you see any other issues, otherwise I'll merge shortly.

Thanks,
John

Issue #707, opened 2022-03-25
Warning: Encountered known unsupported method torch.Tensor.transpose


kairos03:
I got some warning while converting model. 

```
Warning: Encountered known unsupported method torch.Tensor.transpose
```

It seems torch2trt has `torch.transpose` converter but it dose not applied to `torch.Tensor.transpose`.
So, I fix this by only add two line.

I'll make PR to fix this.

PR #708, opened 2022-03-25
Add torch.Tensor.transpose converter

kairos03:
torch.Tensor.transpose converter is missed.
I just add the annotation to link torch.Tensor.transpose and torch.transpose


kairos03:
ref: #707

PR #743, opened 2022-06-07
[Dynamic Batch Sizes] Support dynamic batch sizes via optimization profiles

chaoz-dev:
This PR allows torch2trt to support dynamic batch sizes via optimization profiles. 

During first compilation of the model when given explicit static (batch size) tensor shapes, the subsequently compiled TRT model expects _only_ that input shape during inference, resulting in tensors with smaller batch sizes to be rejected during inference as a result. This PR removes this restriction by using optimization profiles to support dynamic tensor shapes. 

Note that this functionality was previously supported inherently due to the use of implicit batch sizes; with the move to explicit batch sizes (which is necessary for forward compatibility, as implicit batch sizes are now considered deprecated in TRT), we now need to explicitly support this functionality.

Also note that for this functionality to work, the batch size provided during inference **must be smaller than or equal to** the batch size provided during compilation (this is the same functionality as before). 

This works with the following script:
```
  import logging    
  import tensorrt    
  import torch    
  import torch2trt    
      
      
  torch.manual_seed(0)    
      
  DEVICE = torch.device("cuda:0")    
  OLD_SHAPE = (2, 2)    
  NEW_SHAPE = (1, 2)    
      
      
  class Model(torch.nn.Module):    
      def __init__(self):    
          super().__init__()    
      
      def forward(self, a):    
          return a + a    
      
      
  if __name__ == "__main__":    
      old_tensor = torch.randn(OLD_SHAPE, dtype=torch.float32, device=DEVICE)    
      new_tensor = torch.randn(NEW_SHAPE, dtype=torch.float32, device=DEVICE)    
      
      model = Model().eval().to(DEVICE)    
      out = model(old_tensor)    
      print(f'Old: {out}')    
      out = model(new_tensor)    
      print(f'New: {out}')    
      
          model, [old_tensor], max_batch_size=2, log_level=tensorrt.Logger.INFO    
      )    
      out_trt = model_trt(old_tensor)    
      print(f'Old: {out_trt}')    
      out_trt = model_trt(new_tensor)    
      print(f'New: {out_trt}')    
      
      delta = torch.abs(out - out_trt)    
      print(f'Delta: {delta}')    
      assert torch.max(delta) < 1e-6    
```
Outputs:
```
(torch2trt-master-py3.7.7) ~/workspace/pytorch-scratch/torch2trt $ python multi_input.py
Old: tensor([[-1.8493, -0.8507],
        [-5.2877,  0.2904]], device='cuda:0')
New: tensor([[ 0.3617, -1.1045]], device='cuda:0')
[06/07/2022-16:52:34] [TRT] [I] [MemUsageChange] Init CUDA: CPU +349, GPU +0, now: CPU 2986, GPU 1675 (MiB)
[06/07/2022-16:52:35] [TRT] [I] [MemUsageSnapshot] Begin constructing builder kernel library: CPU 3003 MiB, GPU 1675 MiB
[06/07/2022-16:52:35] [TRT] [I] [MemUsageSnapshot] End constructing builder kernel library: CPU 3378 MiB, GPU 1799 MiB
[06/07/2022-16:52:35] [TRT] [W] Tensor DataType is determined at build time for tensors not marked as input or output.
[06/07/2022-16:52:36] [TRT] [W] TensorRT was linked against cuBLAS/cuBLAS LT 11.8.0 but loaded cuBLAS/cuBLAS LT 110.9.2
[06/07/2022-16:52:36] [TRT] [I] [MemUsageChange] Init cuBLAS/cuBLASLt: CPU +873, GPU +378, now: CPU 4251, GPU 2177 (MiB)
[06/07/2022-16:52:36] [TRT] [I] [MemUsageChange] Init cuDNN: CPU +127, GPU +60, now: CPU 4378, GPU 2237 (MiB)
[06/07/2022-16:52:36] [TRT] [I] Local timing cache in use. Profiling results in this builder pass will not be stored.
[06/07/2022-16:52:36] [TRT] [I] Detected 1 inputs and 1 output network tensors.
[06/07/2022-16:52:36] [TRT] [I] Total Host Persistent Memory: 0
[06/07/2022-16:52:36] [TRT] [I] Total Device Persistent Memory: 0
[06/07/2022-16:52:36] [TRT] [I] Total Scratch Memory: 0
[06/07/2022-16:52:36] [TRT] [I] [MemUsageStats] Peak memory usage of TRT CPU/GPU memory allocators: CPU 0 MiB, GPU 4 MiB
[06/07/2022-16:52:36] [TRT] [I] Total Activation Memory: 0
[06/07/2022-16:52:36] [TRT] [I] [MemUsageChange] TensorRT-managed allocation in building engine: CPU +0, GPU +0, now: CPU 0, GPU 0 (MiB)
[06/07/2022-16:52:36] [TRT] [I] [MemUsageChange] TensorRT-managed allocation in IExecutionContext creation: CPU +0, GPU +0, now: CPU 0, GPU 0 (MiB)
Old: tensor([[-1.8493, -0.8507],
        [-5.2877,  0.2904]], device='cuda:0')
New: tensor([[ 0.3617, -1.1045]], device='cuda:0')
Delta: tensor([[0., 0.]], device='cuda:0')
```

### DIFF
--- a/torch2trt/torch2trt.py
+++ b/torch2trt/torch2trt.py
@@ -106,7 +106,7 @@ def torch_dim_to_trt_axes(dim):
     # create axes bitmask for reduce layer
     axes = 0
     for d in dim:
-        axes |= 1 << d 
+        axes |= 1 << d
 
     return axes
 
@@ -344,7 +344,7 @@ def device_type_str(device_type):
         return 'GPU'
     elif device_type == trt.DeviceType.DLA:
         return 'DLA'
-    
+
 
 class NetworkWrapper(object):
     def __init__(self, ctx, network):
@@ -353,7 +353,7 @@ class NetworkWrapper(object):
         self._layer_counts = defaultdict(lambda: 0)
 
     def _configure_layer(self, layer):
-        
+
         # set layer device type
         device_type = self._ctx.current_device_type()
         self._ctx.builder_config.set_device_type(layer, device_type)
@@ -361,7 +361,7 @@ class NetworkWrapper(object):
         if not self._ctx.builder_config.can_run_on_DLA(layer) and device_type == trt.DeviceType.DLA:
             if self._ctx.torch2trt_kwargs['gpu_fallback']:
                 device_type = trt.DeviceType.GPU  # layer will fall back to GPU
-        
+
         # set layer name
         def arg_str(arg):
             if isinstance(arg, torch.Tensor):
@@ -371,14 +371,14 @@ class NetworkWrapper(object):
         self._layer_counts[scope_name] += 1
         args = [arg_str(arg) for arg in self._ctx.method_args]
         kwargs = ["%s=%s" % (key, arg_str(arg)) for key, arg in self._ctx.method_kwargs.items()]
-        layer.name = scope_name + ':' + str(self._layer_counts[scope_name] - 1) + ':' + layer.type.name + ':' + device_type_str(device_type) 
-        
+        layer.name = scope_name + ':' + str(self._layer_counts[scope_name] - 1) + ':' + layer.type.name + ':' + device_type_str(device_type)
+
         if orig_device_type != device_type:
             layer.name = layer.name + '(' + device_type_str(orig_device_type) + ')'
 #         "%s [%s #%d, %s] %s(%s)" % (self._ctx.current_module_name(), layer.type.name, self._layer_counts[layer.type.name], device_type_str(device_type),
 #                                           self._ctx.method_str, ", ".join(args + kwargs))
-    
-        
+
+
     def __getattr__(self, name):
         attr = getattr(self._network, name)
         if callable(attr):
@@ -395,7 +395,7 @@ class NetworkWrapper(object):
 
 
 class ConversionContext(object):
-    
+
     def __init__(self, network, converters=CONVERTERS, torch2trt_kwargs=None, builder_config=None):
         self.network = NetworkWrapper(self, network)
         self.lock = False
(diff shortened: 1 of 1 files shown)


torch2trt/torch2trt.py
```diff
@@ -625,6 +626,12 @@ def torch2trt(module,
     if output_names is None:
         output_names = default_output_names(len(outputs))
 
+    # Create optimization profiles. This allows for dynamic shapes, especially dynamic batch sizes.
+    opt_profile = builder.create_optimization_profile();
+    for input_name, input_shape in zip(input_names, input_shapes):
+        opt_profile.set_shape(input_name, (1,) + input_shape[1:], input_shape, input_shape)
```
chaoz-dev, line 632: I should probably set the max here to be that of the provided `max_batch_size` (which may be greater than the provided input tensor)

jaybdub:
Hi @chaoz-dev ,

Thanks for this PR.  

With explicit shapes, I wonder if it is worth giving the user explicit control over which axes are dynamic.  

```python3
data = torch.zeros(1, 3, 224, 224).cuda().eval()
model_trt = torch2trt(model, [data], input_names=['image'], dynamic_axes={'image': [0]})
```

Or in the case where ``input_names=None``

```python3
model_trt = torch2trt(model, [data], dynamic_axes={'input_0': [0]})
```

using the argument provided to ``input_names``.  This is similar to the ``torch.onnx.export`` API.

And then, perhaps having this default to the "old" behavior of assuming the batch dimension is dynamic, by generating the following when ``dynamic_axes='implicit_batch'``

```python3
def torch2trt(..., dynamic_axes='implicit_batch'):
    ...
    if dynamic_axes == 'implicit_batch':
        dynamic_axes = {}
        for name in input_names:
            dynamic_axes[name] = [0]
    ...
```

Thoughts?

John

jaybdub:
Hi @chaoz-dev ,

I've recently come across need for dynamic shapes to optimize EasyOCR with torch2trt.  I've added support for this in the following branch.

https://github.com/jaybdub/torch2trt/tree/easyocr

This introduces the ability to set ``max_shapes``, ``min_shapes``, and ``opt_shapes`` via torch2trt.  If you don't explicitly provide this information, it will infer it from the data provided.

An example is in this test case for how to use it https://github.com/jaybdub/torch2trt/blob/easyocr/torch2trt/dynamic_shape_test.py.

Does this work for your use case?

Best,
John

chaoz-dev:
Hey @jaybdub!

Apologies for the delayed response! I haven't had a chance to look at your suggestions until recently 😬 

With regards to your suggestions, I think this would work for my particular use case, which really only requires supporting dynamic batch sizes up to a given value (which your `implicit_batch` argument would support). 

That said, I do have a few questions here -- 
1. Is there a reason to not allow dynamic sizes for the entire shape, at all times? Is the concern here that defaulting to dynamic shapes might impact performance? 
2. With regards to `max_shapes`, `min_shapes`, and `opt_shapes`, it sounds like we want to support explicitly setting these values (as opposed to passively deriving these values from `inputs`)? ie. We want different values here than what's provided by `inputs`.
3. With regards to the suggested changes, I do have some concerns that the API is growing more complex:
- we may have some leaky abstractions, by exposing the string arguments as default arguments; these string arguments have specific meaning within the implementation only and will be difficult to change in the future without risking breaking backwards compatibility. 
- there are more implicit relationships between the input arguments, which may get harder to keep track of when there are many many input tensors. eg. For many tensors, we may have long lists of arguments for  `max_shapes`, `min_shapes`, `opt_shapes`, `dynamic_axes`, etc. 
What do you think about the following suggestion:
For `inputs`, we accept either a list of tensors (as currently defined); or, alternatively, a list of containers with the following (example) definition:
```
class InputTensor:
    def __init__(
        self,
        name: Optional[str] = None, # ie. input_name. Default to input_0, input_1, etc. internally when name is not provided.
        dtype: Optional[torch.dtype] = torch.float32, # use with shape arg below to generate input tensor

        # Static option:
        shape: Optional[tuple] = None, # use with dtype above to generate input tensor

        # Dynamic option:
        min_shape: Optional[tuple] = None,
        opt_shape: Optional[tuple] = None,
        max_shape: Optional[tuple] = None,
    ):
    # Encapsulate checks, make sure input arg combinations make sense, etc.
    ....
```
I believe this container should simplify the `torch2trt` interface considerably, while explicitly stating the relationship of the various possible input combinations. This should simplify some of the internal implementation as well, I believe. 

Thoughts? 

Chao

jaybdub:
Hi @chaoz-dev ,

Thanks for the feedback!  You raise lots of good points.

I think some of the points are now addressed in the PR (which differs significantly from my previous comment).

1. No strings 
2. No dynamic_axes argument (implicit when min_shapes != max_shapes)
3. max_batch_size now behaves like legacy (but uses optimization profiles under the hood)

I see what you're saying about the complexity of tracking the associations between different arguments.  In my PR, I'm using min_shapes and max_shapes, and it's working well, but perhaps for models with more complex input schemas it would become difficult.

Perhaps in the future, we could add something like the InputTensor (or InputProfile) and use this to set the default values of min_shapes, max_shapes, opt_shapes when it's found.  That way we can use whichever API is more comfortable for the use case.

Best,
John

jaybdub:
FYI - I've merged the EasyOCR example into master.  This has support for dynamic shapes with the following APIs

```python
module_trt  = torch2trt(module, [data], max_batch_size=4)
```

and

```python
module_trt = torch2trt(module, [data], min_shapes=[(1, 3, 224, 224)], max_shapes=[(4, 3, 224, 224)], opt_shapes=[(2, 3, 224, 224)])
```

Hope this helps for your current use case (dynamic batch size).  I'm still open to discussing simplified / alternate APIs in addition to this, but needed to move forward with this merge for a few reasons.

Thanks again!

Best,
John

chaoz-dev:
Sounds good, thanks! I'll put up a PR with the alternate API that I described here and we can discuss on there what additional changes we would like to see.

I'll close this PR for now due to the other PR having landed.

PR #750, opened 2022-06-21
Fix getitem for Py<3.7

chaoz-dev:


chaoz-dev:
Supplanted by #751 751

PR #764, opened 2022-07-19
[WIP] Easyocr

jaybdub:
- Adds support for dynamic shapes
- Adds dataset classes for shape inference and to simplify calibration

TODO: 

- back-support max_batch_size argument (no action currently)
- handle tuple / dict argument and output types implicitly rather than relying on explicit ignore_inputs argument


jaybdub (commented):
reviewed

light42:
Sorry for kicking a dead-horse. But this easyocr batch inference is basically just stitching N number of images into one and infer it with readtext right? The width and height gets larger but the number of batch is still one.

How about using readtext_batched instead? Which method of batching is better?

I've tried modifying your PR to run readtext_batched and I can't get it worked.

Issue #755, opened 2022-06-30
[Bug/Feature Request] `torch.Tensor.__getitem__` doesn't support list or `torch.Tensor` arguments for indexing/slicing when TRT converting


chaoz-dev:
`torch.Tensor.__getitem__` supports list or `torch.Tensor` inputs for indexing, which appears to not convert correctly currently in `torch2trt`. 

In other words, we'd like to perform the following ops, which currently fails or produces incorrect outputs after `torch2trt` conversion:
```
tensor = torch.randn(2, 3)
tensor[[1, 1, 1]] # index using list
tensor[torch.tensor([1, 1, 1])] # index using tensor
```

This can be demonstrated with the following script:

`getitem-tensors.py`:
```
  import logging    
  import tensorrt    
  import torch    
  import torch2trt    
      
      
  logging.basicConfig(level=logging.INFO)    
  torch.manual_seed(0)    
      
  DEVICE = torch.device("cuda:0")    
  SHAPE = (2, 3)    
      
      
  class ListIndexModel(torch.nn.Module):    
      def __init__(self):    
          super().__init__()    
      
      def forward(self, a):    
          return a[[1, 1, 1, 1]]    
      
      
  class TensorIndexModel(torch.nn.Module):    
      def __init__(self):    
          super().__init__()    
          self.index_tensor = torch.tensor([1, 1, 1, 1]).to(DEVICE)    
      
      def forward(self, a):    
          return a[self.index_tensor]    
      
      
  if __name__ == "__main__":    
      tensor = torch.randn(SHAPE, dtype=torch.float32, device=DEVICE)    
      print(f'Input: {tensor}')    
      
      list_model = ListIndexModel().eval().to(DEVICE)    
      list_out = list_model(tensor)    
      print(f'List index model: {list_out}')    
      
      try:    
          list_model_trt = torch2trt.torch2trt(    
              list_model, [tensor], max_batch_size=SHAPE[0], log_level=tensorrt.Logger.INFO    
          )    
          list_out_trt = list_model_trt(tensor)    
          print(f'TRT List index model: {list_out_trt}')    
      except:    
          print('ERROR: List index model failed to convert.')    
      
      tensor_model = TensorIndexModel().eval().to(DEVICE)    
      tensor_out = tensor_model(tensor)    
      print(f'Tensor index model: {tensor_out}')

      try:    
          tensor_model_trt = torch2trt.torch2trt(    
[...]

Issue #769, opened 2022-07-23
[Question/Bug?] Release v0.4.0 seems to default to input shapes of [1, ...]. Will this work for ops on the batch dim? 


chaoz-dev:
I just saw that v0.4.0 was released a few hours ago. Nice! 

Looking through the changes, I noticed the following code snippet:
```
    # handle inputs as dataset of list of tensors
    if issubclass(inputs.__class__, Dataset):
        dataset = inputs
        if len(dataset) == 0:
            raise ValueError('Dataset must have at least one element to use for inference.')
        inputs = dataset[0]
    else:
        dataset = TensorBatchDataset(inputs)
        inputs = dataset[0]
```

This seems to default `inputs` to shape [1, ....] for all inputs again, as we used to do for implicit batch dimension. 
I believe this was changed with the addition of the explicit batch dimension, prior to the release of v0.4.0? 

Will this change work for inputs where the first dim has a value > 1, and we wish to operate on this dim? 
For example, 
```
t = torch.rand(7, 2, 3, 4)
t[3]
```

Running the following script does seem to imply that we will have issues here:
```
  import logging
  import tensorrt
  import torch
  import torch2trt
  
  
  logging.basicConfig(level=logging.INFO)
  torch.manual_seed(0)
  
  DEVICE = 'cuda:0'
  TENSOR = torch.rand(7, 3, 4).to(DEVICE)
  
  
  class Model(torch.nn.Module):
      def __init__(self):
          super().__init__()
  
      def forward(self, tensor):
          return tensor[3]
  
  
  if __name__ == "__main__":
      model = Model().eval().to(DEVICE)
      out = model(TENSOR)
      print(f'Expected model output: {out}')
  
      model_trt = torch2trt.torch2trt(
          model, [TENSOR], max_batch_size=TENSOR.shape[0], log_level=tensorrt.Logger.INFO
      )
      out = model_trt(TENSOR)
      print(f'TRT model output: {out}')
```
Outputs:
```
(torch2trt-master-py3.7.7) ~/workspace/pytorch-scratch/torch2trt $ python getitem-element.py
Expected model output: tensor([[0.2081, 0.9298, 0.7231, 0.7423],
        [0.5263, 0.2437, 0.5846, 0.0332],
        [0.1387, 0.2422, 0.8155, 0.7932]], device='cuda:0')
Traceback (most recent call last):
  File "getitem-element.py", line 28, in <module>
[...]

PR #770, opened 2022-07-23
[torch2trt/converters] Add element support to `torch.Tensor.__getitem__` converter

### LINKED ISSUE #768: [Bug] `torch.Tensor.__getitem__` fails when indexing/slicing with a single element, with error: `TypeError: 'int' object is not iterable`
**Problem**
The `torch2trt` conversion for `torch.Tensor.__getitem__` fails when indexing/slicing for single element; eg.
```
tensor = torch.rand(2, 3)
tensor[0]
```
The issue is in `torch2trt/converters/getitem.py`, where `slices` is assumed to be iterable. This is not necessarily true in the aforementioned use case, where slices will actually be a single element (specifically the int given as the indexing argument). 

**Script**
Running the following script `getitem-element.py` using NGC `22.06-py3`:
```
  import logging    
  import tensorrt    
  import torch    
  import torch2trt    
      
      
  torch.manual_seed(0)    
      
  DEVICE = 'cuda:0'    
  TENSOR = torch.rand(2, 3).to(DEVICE)    
      
      
  class Model(torch.nn.Module):    
      def __init__(self):    
          super().__init__()    
      
      def forward(self, tensor):    
          return tensor[0]    
      
      
  if __name__ == "__main__":    
      model = Model().eval().to(DEVICE)    
      out = model(TENSOR)    
      print(f'Expected model output: {out}')    
      
          model, [TENSOR], max_batch_size=TENSOR.shape[0], log_level=tensorrt.Logger.INFO    
      )    
      out = model_trt(TENSOR)    
      print(f'TRT model output: {out}')    
```
produces the following output:
```
root@8f319e91dd9a:/opt# python /scripts/getitem-element.py
Expected model output: tensor([0.4963, 0.7682, 0.0885], device='cuda:0')
[07/22/2022-23:49:16] [TRT] [I] [MemUsageChange] Init CUDA: CPU +464, GPU +0, now: CPU 1299, GPU 817 (MiB)
[07/22/2022-23:49:16] [TRT] [I] [MemUsageSnapshot] Begin constructing builder kernel library: CPU 1299 MiB, GPU 817 MiB
[07/22/2022-23:49:16] [TRT] [I] [MemUsageSnapshot] End constructing builder kernel library: CPU 1453 MiB, GPU 859 MiB
Traceback (most recent call last):
  File "/scripts/getitem-element.py", line 27, in <module>
  File "/opt/conda/lib/python3.8/site-packages/torch2trt-0.4.0-py3.8.egg/torch2trt/torch2trt.py", line 736, in torch2trt
    outputs = module(*inputs)
  File "/opt/conda/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1148, in _call_impl
    result = forward_call(*input, **kwargs)
  File "/scripts/getitem-element.py", line 19, in forward
    return tensor[0]
  File "/opt/conda/lib/python3.8/site-packages/torch2trt-0.4.0-py3.8.egg/torch2trt/torch2trt.py", line 307, in wrapper
    converter["converter"](ctx)
  File "/opt/conda/lib/python3.8/site-packages/torch2trt-0.4.0-py3.8.egg/torch2trt/converters/getitem.py", line 34, in convert_tensor_getitem
    num_ellipsis = len(input.shape) - num_slice_types(slices)
  File "/opt/conda/lib/python3.8/site-packages/torch2trt-0.4.0-py3.8.egg/torch2trt/converters/getitem.py", line 18, in num_slice_types
    for s in slices:
TypeError: 'int' object is not iterable
```

chaoz-dev:
Resolves #768.

This PR adds support for single element arguments to `torch.Tensor.__getitem__` converter. 
We convert the input into a tuple if it is not already a tuple, which can occur when the input argument is a single argument. All other arguments should be handled the same (especially multiple arguments, as they are already tuples). 

Single tuple arguments are unpacked correctly still as expected into multiple indexing arguments.
Note that tuples combined with ellipsis or colon arguments are still not yet supported; see #755.

Also, add a lot of unit tests to check that operations on the batch dimension are handled correctly.

### DIFF
--- a/torch2trt/converters/getitem.py
+++ b/torch2trt/converters/getitem.py
@@ -3,17 +3,17 @@ from torch2trt.module_test import add_module_test
 
 
 def slice_to_trt(ctx, dim_size, dim_slice):
-    
+
     start = 0 if dim_slice.start is None else dim_slice.start
     stop = dim_size if dim_slice.stop is None else dim_slice.stop
     stride = 1 if dim_slice.step is None else dim_slice.step
-    
+
     start = make_int_wrapper(start)
     stop = make_int_wrapper(stop)
     stride = make_int_wrapper(stride)
 
     size = (stop - start - 1) // stride + 1
-    
+
     return start, size, stride
 
 
@@ -30,19 +30,33 @@ def convert_tensor_getitem(ctx):
     input = ctx.method_args[0]
     slices = ctx.method_args[1]
     output = ctx.method_return
-    
+
     if not hasattr(input, '_trt'):
         return
 
     input_trt = input._trt
-    
+
+    # Convert slices argument into a tuple if it is not already one.
+    # This can happen when the provided slices argument is a single element.
+    #
+    # In the general case, multiple slice arguments will already be wrapped in a tuple.
+    #
+    # Note the special case where a single tuple is itself the only slice argument (eg. t[(...)]);
+    # PyTorch treats the single tuple argument as if its elements were provided as separate slice arguments.
+    # We handle this gracefully by default without requiring special handling.
+    #
+    # A tuple of tuple of elements is only possible if the inner tuple is combined with other arguments.
+    # In this case, the inner tuple is now a gather operation and must be handled differently.
+    # This is also commonly known as advanced indexing.
+    if not isinstance(slices, tuple):
+        slices = (slices,)
+
     # Step 1 - Replace ellipsis with expanded slices
-    
     num_ellipsis = len(input.shape) - num_slice_types(slices)
-    
+
     new_slices = []
     for s in slices:
-        
+
         if s == Ellipsis:
             while num_ellipsis > 0:
                 new_slices.append(slice(None, None, None))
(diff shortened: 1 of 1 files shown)


chaoz-dev:
Note that this was tested on v0.3.x; the unit tests here fail on v0.4.0. See #769.

chaoz-dev:
Running `python -m torch2trt.test --name getitem`:
```
NUM_TESTS: 39
NUM_SUCCESSFUL_CONVERSION: 39
NUM_FAILED_CONVERSION: 0
NUM_ABOVE_TOLERANCE: 0
NUM_pSNR_TOLERANCE: 0
```

torch2trt/converters/getitem.py
```diff
@@ -117,30 +130,210 @@ def convert_tensor_getitem(ctx):
...
+@add_module_test(torch.float32, torch.device('cuda'), [(3, 2, 4)], max_batch_size=3)
+def test_tensor_getitem_0d_2tuple():
+    return LambdaModule(lambda x: x[(0, 1)])
+
+
+# There is currently an issue with this test case.
+# Need to investigate this more.
+#  @add_module_test(torch.float32, torch.device('cuda'), [(3, 2, 4)], max_batch_size=3)
+#  def test_tensor_getitem_0d_3tuple():
+    #  return LambdaModule(lambda x: x[(0, 1, 2)])
```
chaoz-dev, line 196: @jaybdub Looks like there's an issue with this test case, which we'll need to look into more. I don't _think_ it's related to the changes in this PR, however, but with how we're indexing in general.

chaoz-dev:
```
NUM_TESTS: 38
NUM_SUCCESSFUL_CONVERSION: 38
NUM_FAILED_CONVERSION: 0
NUM_ABOVE_TOLERANCE: 0
NUM_pSNR_TOLERANCE: 0
```

PR #772, opened 2022-07-25
Enhance the einsum converter

LongerZrLong:
Hi there, I notice that the einsum converter can only handle the case: torch.einsum(eq, x, y) while not the case: torch.einsum(eq, [x, y]).


jaybdub:
Hi @LongerZrLong ,

Thanks for the PR!  Looks like a straightforward fix to me.

Will test and likely merge soon.

Best,
John

PR #779, opened 2022-07-29
[WIP] fix various converters for dynamic shapes

jaybdub:
- adds shape tracking with wrapper classes

Fixed:

- [x] flatten
- [x] squeeze
- [x] unsqueeze
- [x] view
- [x] reshape
- [x] interpolate
- [x] __getitem__
- [ ] chunk
- [ ] split

Chunk / split can return variable number of outputs for dynamic shapes.  They currently don't work with ONNX in this context.  A good default behavior in this instance needs to be defined, but for now can probably be omitted as not supported by dynamic shapes.


torch2trt/torch2trt.py
```diff
@@ -865,3 +892,164 @@ def set_layer_precision(ctx, layer):
...
+_int_sub = int.__sub__
+_int_floordiv = int.__floordiv__
+
+class IntWrapper(int):
+    
+    @property
+    def _trt(self):
+        if not hasattr(self, '_raw_trt'):
+            ctx = get_conversion_context()
+            self._raw_trt = ctx.network._network.add_constant([1], np.array([_int(self)], dtype=np.int32)).get_output(0)
```
chaoz-dev, line 920: I've been wondering about how this works... Normally, we would detach the value here when creating a constant TRT value, but in this case it looks like we're just passing the wrapper directly into the (numpy array) constant layer instead? Since we use this class primarily for shape values, when the shapes change, does this TRT value change because the underlying int value is changing?

jaybdub: This adds the _raw_trt value if it doesn't already exists.

However, _raw_trt is set 

1. When the integer is created as part of a tensor shape (see _shape_wrapper line 1041 below)
2. When the integer is the result of arithmetic of other integers (see operations below)

jaybdub:
@chaoz-dev @SrivastavaKshitij 

I've addressed a few bugs in this PR.  Now all test cases are passing.

I believe this should be good to merge, please let me know if this is still working for your use case.

torch2trt/torch2trt.py
chaoz-dev: Ah okay after stepping through the code some more, I think I understand how this progression works now.. 
I understood how the `_raw_trt` was getting set before, but was confused as to how this was tracking the shape changes as the model was being fed differently shaped inputs. (In the updated PR) it turns out this was just getting set here:
```
      shape_trt = ctx.network._network.add_shape(input._trt).get_output(0)

      for i, d in enumerate(output):
          d._raw_trt = ctx.network._network.add_slice(shape_trt, [i], [1], [1]).get_output(0)
```
Specifically, we just set the `_raw_trt` for each dimension to be the output of the shape layer, which will correctly track the shape of the tensor regardless of the input tensor shape.

torch2trt/torch2trt.py
```diff
@@ -865,3 +896,177 @@ def set_layer_precision(ctx, layer):
...
+    stack = [True] # default true
+
+    def __init__(self, value: bool):
+        self._value = value
+    
+    def __enter__(self, *args, **kwargs):
+        self.stack.insert(0, self._value)
+
+    def __exit__(self, *args, **kwargs):
+        self.stack.pop()
```
chaoz-dev, line 1072: What's the reason for tracking the shape wrapping here? I see this is called in several of the helper functions. I'm guessing it's to prevent calling `_size_wrapper` within nested (or recursive) calls to the helper functions, but I don't think I've fully grokked the use case yet. 

Also, should this actually be renamed to a queue? It looks like we're inserting to the front but popping from the back.

chaoz-dev:
Stepped through the code just to make sure that I understand the changes and to see the ops running with dynamic shapes, and it looks good to me as far as functionality goes. 

I did find one issue, however, but I'm not quite sure where it's coming from yet... Observe the following:
Running this script `dynamic-views.py`:
```
  import logging
  import tensorrt
  import torch
  import torch2trt


  torch.manual_seed(0)

  DEVICE = torch.device("cuda:0")


  def create_model(model_type, tensors):
      shapes = [tensor.shape for tensor in tensors]

      model = model_type().eval().to(DEVICE)
          model,
          tensors,
          max_batch_size=shapes[0][0],
          #  log_level=tensorrt.Logger.INFO,
          min_shapes = [(1,) * len(shape) for shape in shapes],
          max_shapes = shapes,
          opt_shapes = shapes,
      )

      return model, model_trt


  def to_tensor(shape):
      return torch.rand(shape).to(DEVICE)


  def run_model(model, model_trt, tensors):
      out = model(*tensors)
      out_trt = model_trt(*tensors)

      assert torch.allclose(out, out_trt, atol=1e-4, rtol=1e-4)
      print(f"Successfully run model {type(model)} with input shapes {[tensor.shape for tensor in tensors]} and out shape {out.shape} out_trt shape {out_trt.shape}")


  class AddModel(torch.nn.Module):
      def forward(self, a):
          return a + a


  class ViewModel(torch.nn.Module):
      def forward(self, a):
          shape = a.shape
          return a.view(shape)

  def run_dynamic(model_type):
      shapes = [(5, 5, 5), (3, 3, 3), (1, 1, 1)]
      t = to_tensor(shapes[0])
      model, model_trt = create_model(model_type, [t])

      for shape in shapes:
          t = to_tensor(shape)
          run_model(model, model_trt, [t])


  if __name__ == "__main__":
      run_dynamic(ViewModel)
      run_dynamic(AddModel)
      run_dynamic(ViewModel)
```
we see the following output:
```
$ python dynamic-views.py
Successfully run model <class '__main__.ViewModel'> with input shapes [torch.Size([5, 5, 5])] and out shape torch.Size([5, 5, 5]) out_trt shape torch.Size([5, 5, 5])
Successfully run model <class '__main__.ViewModel'> with input shapes [torch.Size([3, 3, 3])] and out shape torch.Size([3, 3, 3]) out_trt shape torch.Size([3, 3, 3])
Successfully run model <class '__main__.ViewModel'> with input shapes [torch.Size([1, 1, 1])] and out shape torch.Size([1, 1, 1]) out_trt shape torch.Size([1, 1, 1])
Successfully run model <class '__main__.AddModel'> with input shapes [torch.Size([5, 5, 5])] and out shape torch.Size([5, 5, 5]) out_trt shape torch.Size([5, 5, 5])
Successfully run model <class '__main__.AddModel'> with input shapes [torch.Size([3, 3, 3])] and out shape torch.Size([3, 3, 3]) out_trt shape torch.Size([3, 3, 3])
Successfully run model <class '__main__.AddModel'> with input shapes [torch.Size([1, 1, 1])] and out shape torch.Size([1, 1, 1]) out_trt shape torch.Size([1, 1, 1])
Successfully run model <class '__main__.ViewModel'> with input shapes [torch.Size([5, 5, 5])] and out shape torch.Size([5, 5, 5]) out_trt shape torch.Size([5, 5, 5])
[10/06/2022-18:31:50] [TRT] [E] 7: [shapeMachine.cpp::execute::702] Error Code 7: Internal Error (IShuffleLayer :0:SHUFFLE:GPU: reshaping failed for tensor: input_0
reshape would change volume
Instruction: RESHAPE_ZERO_IS_PLACEHOLDER{3 3 3} {5 5 5}
)
Traceback (most recent call last):
  File "dynamic-views.py", line 66, in <module>
    run_dynamic(ViewModel)
  File "dynamic-views.py", line 60, in run_dynamic
    run_model(model, model_trt, [t])
  File "dynamic-views.py", line 36, in run_model
    out_trt = model_trt(*tensors)
  File "/home/chaoz/.anaconda3/envs/torch2trt-master-py3.7.7/lib/python3.7/site-packages/torch/nn/modules/module.py", line 1110, in _call_impl
    return forward_call(*input, **kwargs)
  File "/home/chaoz/.anaconda3/envs/torch2trt-master-py3.7.7/lib/python3.7/site-packages/torch2trt-0.4.0-py3.7.egg/torch2trt/torch2trt.py", line 623, in forward
    shape = tuple(self.context.get_binding_shape(idx))
ValueError: __len__() should return >= 0
```

In `dynamic-views.py`, we're creating three trivial models (`ViewModel`, `AddModel`, then `ViewModel` again) and setting them to run with dynamic shapes (max length 5, min length 1, in each of three dimensions). 

For some reason, we run into the following error:
```
[10/06/2022-18:31:50] [TRT] [E] 7: [shapeMachine.cpp::execute::702] Error Code 7: Internal Error (IShuffleLayer :0:SHUFFLE:GPU: reshaping failed for tensor: input_0
reshape would change volume
Instruction: RESHAPE_ZERO_IS_PLACEHOLDER{3 3 3} {5 5 5}
)
```
when attempting to torch2trt convert the `ViewModel` a second time, after converting and running the `AddModel`. Note that the first conversion and running of the `ViewModel` succeeds.

torch2trt/converters/view.py
```diff
@@ -1,21 +1,23 @@
...
 @tensorrt_converter('torch.Tensor.view')
-@tensorrt_converter('torch.Tensor.squeeze')
-@tensorrt_converter('torch.Tensor.unsqueeze')
-@tensorrt_converter('torch.Tensor.flatten')
-@tensorrt_converter('torch.squeeze')
-@tensorrt_converter('torch.unsqueeze')
+@tensorrt_converter('torch.Tensor.reshape')
 def convert_view(ctx):
     input = ctx.method_args[0]
-    input_trt = add_missing_trt_tensors(ctx.network, [input])[0]
```
chaoz-dev, line 15: Is this okay to remove? I'm currently running into a situation where I need to attach a TRT tensor to the input because it doesn't already exist...

chaoz-dev: Just did some testing, looks like we need to call this and similar methods because member variable tensors of the class do not automatically get converted to TRT (ie. gets the `._trt` attribute) until they are encountered in a converter/operation. Whereas most tensors get this attribute by default on the output of a converter (ie. `output._trt`), this is not the case for member variables because their first encounter is in the op in which they are used, at which point they might yet be initialized with a `._trt` attribute.

chaoz-dev, line 15: `input._trt = add_missing_trt_tensors(ctx.network, [input])[0]`

jaybdub: Hi @chaoz-dev ,

I ran into this issue as well while optimizing timm's maxvit model.

I addressed this in the following branch by deferring adding constants until a non-unary operation is found that has at least one non-constant trt input.

This is done in this branch, and may address this issue for most converters.

https://github.com/NVIDIA-AI-IOT/torch2trt/compare/master...jaybdub:torch2trt:maxvit

jaybdub: I'll need to test, but I can re-base this PR on these changes.

torch2trt/torch2trt.py
jaybdub: The purpose is like you describe.  For utility functions we want to access the shape of the torch tensor without modifying the TensorRT network (like in utility functions).

This context manager enables / disables the shape tracking.

This should be a stack, that's a bug! Insert should be append.  I'll update this and test.

jaybdub:
I've re-based this change on the branch that fixes the add_missing_trt_tensors issue, and fixed the stack implementation.

Please let me know if this works for you and I can merge quickly.  I've checked that test cases are passing.

chaoz-dev:
Awesome thanks! Yeah I can confirm that adding 
```
input._trt = add_missing_trt_tensors(ctx.network, [input])[0]
```
back in resolves these issues... I've actually been running with this patch locally for a while now and have not had any further issues with this problem.

chaoz-dev:
Ah awesome! I'll pull this latest change into our repo and test if everything is still working as they are currently. So far I've just been running with `input._trt = add_missing_trt_tensors(ctx.network, [input])[0]` patched in, but if the changes still behave the same as this patch then I think we should be okay...

chaoz-dev:
Quick update on my side -- Still testing the newest changes (from this PR only), but my initial tests show that the above issues regarding `dynamic-views.py` appear to be resolved. 

I tested a toy example with regards to the `add_missing_trt_tensors` issue and that seems okay for now as well. I had some concerns regarding deferring adding constant tensors, as we may need this functionality for tensors that are found as member variables for a module, but in practice this may still be okay.

chaoz-dev:
All of the model exports on my side have completed and metrics look good (ie. no-op) so we're good on our side 👍🏼

PR #800, opened 2022-08-31
Fix issues when using interpolate in versions of python >= 3.8

mahdilamb:
collections.Sequence was moved to collections.abc.Sequence. This is a fix so that it can work in versions before and after python 3.8


stap-odin:
This is still relevant, I came to do the same thing, can someone merge this ?

PR #804, opened 2022-09-13
QAT : TRT 8 compatible workflow

SrivastavaKshitij:
Hi @jaybdub 

I am introducing this new QAT workflow which is compatible with TensorRT 8. 

TRT introduced IQuantize and IDequantize Layers which are to be manually placed in the network based on the guidelines mentioned in [Q/DQ placement](https://docs.nvidia.com/deeplearning/tensorrt/developer-guide/index.html#qdq-placement-recs).  

I have added support to quantize `nn.Conv2d`, `nn.MaxPool2d` and `nn.AdaptiveAvfPool2d` - layers that are necessary to quantize Resnet(s). I have also added a `QuantGenericTensor` which can be used to add QDQ layer anywhere in the model based on Nvidia's guidelines. 

This PR also introduces the option to choose between per tensor quantization and per channel quantization. All quant layers are scriptable with `torch.jit.script`

Most of the files that I have modified / changed are under `contrib` folders, so it doesn't affect the main torch2trt library. 

I will continue to add support for more layers but I believe this PR is big enough to land and then I can put up smaller PRs to add more functionalities. 

Entire workflow is tested with Pytorch NGC Container `22.04-py3`

Thanks.


torch2trt/contrib/qat/converters/QuantAdaptiveAvgPool2d.py
```diff
@@ -0,0 +1,45 @@
...
+    if hasattr(module._input_quantizer,'quant_axis'):
+        input_quantizer.axis = module._input_quantizer.quant_axis.to(torch.long).item()
+    else:
+        input_quantizer.axis=0
+
+    input_dequantizer = ctx.network.add_dequantize(
+            input = input_quantizer.get_output(0),
+            scale = scale_trt.get_output(0))
+
+    if hasattr(module._input_quantizer,'quant_axis'):
```
praveen-aurora, line 34: This seems like it can be simplified by re-using the result from the if block at line 25

Issue #677, opened 2021-12-17
IndexError: invalid index of a 0-dim tensor. Use `tensor.item()` in Python or `tensor.item<T>()` in C++ to convert a 0-dim tensor to a number (base) [root@A09-R34-I192-88-FZ6JTP2 fastspeech]# git branch


weixsong:
Hi,

In my model, I used a dynamic mask from input length:
```
def get_mask_from_lengths(lengths, max_len=None):
    batch_size = lengths.shape[0]
    if max_len is None:
        max_len = torch.max(lengths)

    ids = torch.arange(0, max_len).unsqueeze(0).expand(batch_size, -1).to(device)
    mask = ids >= lengths.unsqueeze(1).expand(-1, max_len)

    return mask
```

but when I tried to export tensorrt model by torch2trt, I meet the following error:
```
Traceback (most recent call last):
  File "torch2trt_tutorial.py", line 207, in <module>
    model_trt = torch2trt(model, [speakers, texts, src_lens, duration_control])
  File "/export/anaconda3/lib/python3.7/site-packages/torch2trt-0.3.0-py3.7.egg/torch2trt/torch2trt.py", line 553, in torch2trt
    outputs = module(*inputs)
  File "/export/anaconda3/lib/python3.7/site-packages/torch/nn/modules/module.py", line 1102, in _call_impl
    return forward_call(*input, **kwargs)
  File "/export/users/songwei/workspace/FastSpeech/exp02_test/fastspeech/model/fastspeech2.py", line 35, in forward
    src_masks = get_mask_from_lengths(src_lens, max_src_len)
  File "/export/users/songwei/workspace/FastSpeech/exp02_test/fastspeech/utils/tools.py", line 63, in get_mask_from_lengths
    max_len = torch.max(lengths).item()
  File "/export/anaconda3/lib/python3.7/site-packages/torch2trt-0.3.0-py3.7.egg/torch2trt/torch2trt.py", line 300, in wrapper
    converter["converter"](ctx)
  File "/export/anaconda3/lib/python3.7/site-packages/torch2trt-0.3.0-py3.7.egg/torch2trt/converters/max.py", line 33, in convert_max
    __convert_max_reduce(ctx)
  File "/export/anaconda3/lib/python3.7/site-packages/torch2trt-0.3.0-py3.7.egg/torch2trt/converters/max.py", line 21, in __convert_max_reduce
    output_val = ctx.method_return[0]
[...]

PR #826, opened 2022-12-15
[torch2trt/converters] Add support for single tensor only input arguments for `torch.max` and `torch.min`

chaoz-dev:
**Description** 
This PR resolves the issue where `torch.max` and `torch.min` do not correctly convert if they're given only a single tensor input, which is a common use case. 

Addresses issue https://github.com/NVIDIA-AI-IOT/torch2trt/issues/677

**Tests**
See additional unit tests for `torch.max` and `torch.min`:
```
|                                      torch2trt.converters.max.test_max | float32 |               [(3, 3, 3)] | {} | 0.00E+00 | nan | 0.00E+00 | 7.72e+04 | 1.51e+04 | 0.0301 | 0.0838 |
|                                      torch2trt.converters.max.test_max | float32 |               [(1, 3, 3)] | {} | 0.00E+00 | nan | 0.00E+00 | 7.81e+04 | 1.52e+04 | 0.03 | 0.0831 |
|                                      torch2trt.converters.max.test_max | float32 |                  [(1, 3)] | {} | 0.00E+00 | nan | 0.00E+00 | 7.8e+04 | 1.52e+04 | 0.0299 | 0.0831 |
...
NUM_TESTS: 19
NUM_SUCCESSFUL_CONVERSION: 19
NUM_FAILED_CONVERSION: 0
NUM_ABOVE_TOLERANCE: 0
NUM_pSNR_TOLERANCE: 0
```
```
~/workspace/torch2trt (chaoz/torch-max-min-single-tensor) $ python -m torch2trt.test --name test_min
|                                      torch2trt.converters.min.test_min | float32 |               [(3, 3, 3)] | {} | 0.00E+00 | nan | 0.00E+00 | 7.84e+04 | 1.51e+04 | 0.0298 | 0.0839 |
|                                      torch2trt.converters.min.test_min | float32 |               [(1, 3, 3)] | {} | 0.00E+00 | nan | 0.00E+00 | 7.84e+04 | 1.51e+04 | 0.0296 | 0.0831 |
|                                      torch2trt.converters.min.test_min | float32 |                  [(1, 3)] | {} | 0.00E+00 | nan | 0.00E+00 | 7.9e+04 | 1.51e+04 | 0.0298 | 0.0838 |
...
NUM_TESTS: 11
NUM_SUCCESSFUL_CONVERSION: 11
NUM_FAILED_CONVERSION: 0
NUM_ABOVE_TOLERANCE: 0
NUM_pSNR_TOLERANCE: 0
```

### DIFF
--- a/torch2trt/converters/max.py
+++ b/torch2trt/converters/max.py
@@ -11,18 +11,21 @@ def __convert_max_elementwise(ctx):
     input_a_trt, input_b_trt = broadcast_trt_tensors(ctx.network, [input_a_trt, input_b_trt], len(output.shape))
     layer = ctx.network.add_elementwise(input_a_trt, input_b_trt, trt.ElementWiseOperation.MAX)
     output._trt = layer.get_output(0)
-    
+
 
 def __convert_max_reduce(ctx):
     input = ctx.method_args[0]
     dim = get_arg(ctx, 'dim', pos=1, default=tuple(range(1, len(input.shape))))
     keepdim = get_arg(ctx, 'keepdim', pos=2, default=False)
     input_trt = add_missing_trt_tensors(ctx.network, [input])[0]
-    output_val = ctx.method_return[0]
-    output_idx = ctx.method_return[1]
+    if isinstance(ctx.method_return, torch.Tensor):
+        output_val = ctx.method_return
+    else:
+        output_val = ctx.method_return[0]
+        output_idx = ctx.method_return[1]
     layer = ctx.network.add_reduce(input_trt,  trt.ReduceOperation.MAX, torch_dim_to_trt_axes(dim), keepdim)
     output_val._trt = layer.get_output(0)
-    
+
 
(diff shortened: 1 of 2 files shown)


chaoz-dev:
Just discovered an issue here that the current implementation (in general, not just this PR) does not cover -- because we assumed implicit batch dimensions before, all of the test cases use (1, ...) as the batch size, and `dims` is a tuple that starts at dim one. However, with explicit batch sizes that we now use, we need to update the implementation so that `dims` starts at 0 instead, so all dims are reduced on. Note that this will produce a 0 dim tensor, which I _believe_ is valid in the network but not at the entry/exit of the network; to get around this for unit testing, I think we can unsqueeze before exiting the network so that there is one dimension.

chaoz-dev:
Okay I think this issue is fixed as well

jaybdub:
LGTM

PR #884, opened 2023-10-13
Update interpolate.py: Adds support for Python version >= 3.10

teodorToshkov:
Python deprecated collections.Sequence in favor of collections.abc.Sequence. The update checks for the Python version and uses collections.abc instead of collections for Python versions over 3.9


jaybdub:
Thank you for the fix!

